PR #44, opened 2025-09-29
Project level command management

timothyF95:
## Description

This PR introduces significant changes to how settings and workflow execution are handled in the CLI.

## Key Changes

- **Settings file**  
  - The path to the settings file can no longer be defined directly.  
  - A new `--project-root` flag has been added for headless execution. This allows commands to be executed from anywhere outside of the project.

- **Config flag**  
  - The `--config` flag has been removed.  
  - The config path is defined in the `workflow.yaml` and acts as the single source of truth. If a value is defined in `workflow.yaml`, it will be used automatically.

- **Workflow execution**  
  - You can now execute any command from anywhere within the project, including account linking commands.  
  - Workflow commands take a **positional argument**, which represents the directory path of the workflow.  
    - This path can be absolute or relative to the project root.  
    - Example:  
      ```bash
      cre workflow deploy ./workflows/my-workflow01
      ```  
      This will locate the corresponding `workflow.yaml` in the given directory and execute against that workflow.

### DIFF
--- a/test/multi_command_flows/secrets_happy_path.go
+++ b/test/multi_command_flows/secrets_happy_path.go
@@ -138,8 +138,7 @@ func RunSecretsHappyPath(t *testing.T, tc TestConfig, chainName string) {
 		if !allowed {
 			t.Fatalf("allowlist not detected for create.\n\nCLI OUTPUT:\n%s", out)
 		}
-		require.Contains(t, out, "secret created", "expected create log.\nCLI OUTPUT:\n%s", out)
-		require.Contains(t, out, "success=true", "create should not fail.\nCLI OUTPUT:\n%s", out)
+		require.Contains(t, out, "Secret created:", "expected create log.\nCLI OUTPUT:\n%s", out)
 	})
 
 	// ===== PHASE 2: UPDATE SECRETS =====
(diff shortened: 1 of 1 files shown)


cmd/workflow/deploy/deploy.go
```diff
@@ -73,14 +73,17 @@ var defaultOutputPath = "./binary.wasm.br.b64"
 
 func New(runtimeContext *runtime.Context) *cobra.Command {
 	var deployCmd = &cobra.Command{
-		Use:   "deploy",
+		Use:   "deploy <workflow-folder-path>",
```
shileiwill, line 76: how do this workflow-path and project-root work together? say, if the project-root is an absolute path, and this workflow-folder-path is also absolute, how we ensure the two paths are valid, and workflow path is part of project root?

internal/context/project_context.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/smartcontractkit/cre-cli/internal/constants"
+	"github.com/smartcontractkit/cre-cli/internal/transformation"
+)
+
+// SetExecutionContext sets the appropriate execution context for commands
+// It first sets the project context, then if it's a workflow command with exactly one argument,
+// it changes to the specific workflow directory
+func SetExecutionContext(cmd *cobra.Command, args []string, projectRootFlag string, logger *zerolog.Logger) error {
```
shileiwill, line 18: lets unit test this project_context and workflow_context if it is not too much a burden?

cmd/creinit/creinit.go
```diff
@@ -38,19 +38,20 @@ type WorkflowTemplate struct {
 	Folder string
 	Title  string
 	ID     uint32
-	Lang   TemplateLanguage
 }
 
 type LanguageTemplate struct {
 	Title     string
 	Lang      TemplateLanguage
+	Path      string
```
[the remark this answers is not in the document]
timothyF95: done

internal/context/project_context.go
timothyF95: Added. workflow_context.go is no longer used and has been deleted

ejacquier:
@timothyF95 It is still useful to support passing config value inline with the command execution, so that user can "generate dynamic targets" without bothering to update the target file:

`cre workflow deploy --workflow-path=="./my-wk/main.go" --config-path: "./my-wk/config.json"`

shileiwill (approved):
LG

PR #57, opened 2025-10-02
added secrets path and cleaned up settings templates

timothyF95:


internal/settings/template/workflow.yaml.tpl
```diff
@@ -1,21 +1,18 @@
...
 # my-target:
-#   user-workflow:   
-#     # Required: The name of the workflow to register with the Workflow Registry contract.
-#     workflow-name: "MyExampleWorkflow"
+#   user-workflow:
+#     workflow-name: "MyExampleWorkflow"    # Required: Workflow Registry name
+#   workflow-artifacts:
+#     workflow-path: "./main.ts"            # Path to workflow entry point
+#     config-path: "./config.yaml"          # Path to config file
+#     secrets-path: "../secrets.yaml"       # Path to secrets file (project root by default)
```
shileiwill, line 15: we discussed this yesterday, putting it here means secrets is part of workflow. in fact, secrets operations is workflow agnostic. 

I think keeping it here is neater tho. cc @ejacquier

PR #58, opened 2025-10-02
migrate workflow config to chain name

timothyF95:
Deployed workflow with new config: 006aa0417272d44d7570b9a04841ff41c4f402b543d3235f29d89f50e203e4d9


shileiwill (approved):
I dont see any blockers of getting this change in while we wait for confirmation on using chain name to get evmClient. when that is possible, it would be a quick followup for us to tune the template.

PR #72, opened 2025-10-07
feat: gas cost estimation

anirudhwarrier:


cmd/client/tx.go
```diff
@@ -137,6 +152,20 @@ func (c *TxClient) executeTransactionByTxType(txFn func(opts *bind.TransactOpts)
...
+		if gasErr == nil {
+			gasPrice, gasPriceErr := c.EthClient.Client.SuggestGasPrice(c.EthClient.Context)
+			if gasPriceErr != nil {
+				c.Logger.Warn().Err(gasPriceErr).Msg("Failed to fetch gas price")
+			} else {
+				totalCost := new(big.Int).Mul(new(big.Int).SetUint64(estimatedGas), gasPrice)
+				// Convert from wei to ether for display
+				etherValue := new(big.Float).Quo(new(big.Float).SetInt(totalCost), big.NewFloat(1e18))
+				fmt.Println("Estimated Cost:")
+				fmt.Printf("  Gas Price:      %s wei\n", gasPrice.String())
```
shileiwill, line 165: how about either replace wei with gwei or just use gwei? gwei is more commonly used for gas.

anirudhwarrier: Good callout. Makes sense to show as gwei. https://github.com/smartcontractkit/cre-cli/pull/72/commits/19189a2bd56284fe7451d28bb1f7fc17b761ae15

shileiwill (approved):
LG

PR #89, opened 2025-10-10
Use latest digest creation, and ask for request ID on the second run of MSIG flow

shileiwill:
- latest digest creation is by the jsonrpc.Request's default Digest(): https://github.com/smartcontractkit/chainlink/commit/9e56563c3b1eef632a2baf8351e00827943421d3#diff-0636ba1e7ac59f2197fb75c73618f409938dfcc0fc3536db9546e6b4a292c79a 

- since request ID (UUID) is part of the digest, we need to print the request ID to users and let users copy the UUID to commandline on the second run.

### DIFF
--- a/cmd/secrets/common/handler.go
+++ b/cmd/secrets/common/handler.go
@@ -340,13 +340,15 @@ func (h *Handler) Execute(
 
 	// Resolve request ID (reuse if provided)
 	var requestID string
-	if strings.TrimSpace(requestIDFlag) != "" {
-		if _, err := uuid.Parse(strings.TrimSpace(requestIDFlag)); err != nil {
+	rid := strings.TrimSpace(requestIDFlag)
+	if rid == "" {
+		requestID = uuid.New().String()
+	} else {
+		id, err := uuid.Parse(rid)
+		if err != nil {
 			return fmt.Errorf("--request-id must be a valid UUID: %w", err)
 		}
-		requestID = strings.TrimSpace(requestIDFlag)
-	} else {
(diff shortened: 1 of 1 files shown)


cmd/secrets/create/create.go
```diff
@@ -42,24 +42,28 @@ func New(ctx *runtime.Context) *cobra.Command {
 					Dur("timeout", duration).
 					Dur("maxDuration", maxDuration).
 					Msg(fmt.Sprintf("invalid timeout: must be > 0 and < %dh (%dd)", maxHours, maxDays))
-
 				return fmt.Errorf("invalid --timeout: must be greater than 0 and less than %dh (%dd)", maxHours, maxDays)
 			}
 
-			inputs, err := h.ResolveInputs()
+			reqIDFlag, err := cmd.Flags().GetString("request-id")
```
timothyF95: nit: this feels strangely placed for me. Should it not be part of resolving the inputs?

cmd/secrets/list/list.go
```diff
@@ -51,24 +52,52 @@ func New(ctx *runtime.Context) *cobra.Command {
 				return fmt.Errorf("invalid --timeout: must be greater than 0 and less than %dh (%dd)", maxHours, maxDays)
 			}
 
-			return Execute(h, namespace, duration, ctx.Settings.Workflow.UserWorkflowSettings.WorkflowOwnerType)
+			reqIDFlag, err := cmd.Flags().GetString("request-id")
```
timothyF95: q: How come we don't use viper?

PR #100, opened 2025-10-17
update msig flow for secrets module

shileiwill:
2 steps on MSIG for all 4 operations:
step 1: create a proposal and save the proposal to local

```
leishi@MB-MJ2FP47WKD test98 % cre secrets delete  secrets2.yaml --unsigned

MSIG transaction prepared!

Next steps:

   1. Submit the following transaction on the target chain:
      Chain:            ethereum-testnet-sepolia
      Contract Address: 0xaE55eB3EDAc48a1163EE2cbb1205bE1e90Ea1135

   2. Use the following transaction data:

      94ea0da6fb65a9da4414f94fc78c4d12363b1339bfb40245829cc46acc05cbb4147616960000000000000000000000000000000000000000000000000000000068f47bab

   3. Save this bundle file; you will need it on the second run:
      Bundle Path: fb65a9da4414f94fc78c4d12363b1339bfb40245829cc46acc05cbb414761696.json
      Digest:      0xfb65a9da4414f94fc78c4d12363b1339bfb40245829cc46acc05cbb414761696

   4. After the transaction is finalized on-chain, run:

      cre secrets execute fb65a9da4414f94fc78c4d12363b1339bfb40245829cc46acc05cbb414761696.json

```

step 2: execute the proposal once the onchain allowlist is finalized

```
leishi@MB-MJ2FP47WKD test98 %  cre secrets execute fb65a9da4414f94fc78c4d12363b1339bfb40245829cc46acc05cbb414761696.json
Secret deleted: secret_id=MY_SECRET_ONE1, owner=0x622979Ed50314536f1fC975E10c01Ea984b93d2f, namespace=main
```

### DIFF
--- a/internal/settings/settings_get_test.go
+++ b/internal/settings/settings_get_test.go
@@ -14,17 +14,6 @@ func TestGetWorkflowOwner(t *testing.T) {
 	validPrivKey := "ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
 	expectedOwner := "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
 
-	t.Run("returns owner from flag when set", func(t *testing.T) {
-		v := viper.New()
-		v.Set(settings.CreTargetEnvVar, "test")
-		v.Set(settings.Flags.Owner.Name, "0x123")
-
-		owner, ownerType, err := settings.GetWorkflowOwner(v)
-		assert.NoError(t, err)
-		assert.Equal(t, "0x123", owner)
-		assert.Equal(t, constants.WorkflowOwnerTypeMSIG, ownerType)
-	})
-
 	t.Run("derives owner from eth private key", func(t *testing.T) {
 		v := viper.New()
 		v.Set(settings.CreTargetEnvVar, "test")


cmd/secrets/create/create.go
```diff
@@ -18,7 +18,7 @@ import (
 func New(ctx *runtime.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "create [SECRETS_FILE_PATH]",
-		Short:   "Creates secrets from a YAML file provided as a positional argument.",
+		Short:   "Creates secrets from a YAML file (EOA or MSIG step 1).",
```
timothyF95: As these are the only two options, I'm not sure we need to specify them in the desc

internal/settings/settings_get.go
```diff
@@ -79,10 +81,39 @@ func GetWorkflowOwner(v *viper.Viper) (ownerAddress string, ownerType string, er
 		return "", "", err
 	}
 
+	// if --unsigned flag is set, owner must be set in settings
 	if v.IsSet(Flags.RawTxFlag.Name) {
-		return v.GetString(fmt.Sprintf("%s.%s", target, WorkflowOwnerSettingName)), constants.WorkflowOwnerTypeMSIG, nil
+		key := fmt.Sprintf("%s.%s", target, WorkflowOwnerSettingName)
+
+		if v.IsSet(key) {
+			owner := strings.TrimSpace(v.GetString(key))
```
timothyF95: Validation is done downstream at the command level using go validators: Example: https://github.com/smartcontractkit/cre-cli/blob/16012b74b11b838ef5d6e2aec9c4bb9106f0fb3e/cmd/workflow/deploy/deploy.go#L27 which uses: https://github.com/smartcontractkit/cre-cli/blob/16012b74b11b838ef5d6e2aec9c4bb9106f0fb3e/internal/validation/workflow.go#L20

timothyF95 (changes requested):
Looks mostly good, some small changes. The update to getOwnerAddress needs to return to it's original functionality

cmd/secrets/create/create.go
shileiwill: they are the only 2 options. Good call on should we specify it. I can remove them if it causes confusion.

internal/settings/settings_get.go
```diff
@@ -79,10 +81,39 @@ func GetWorkflowOwner(v *viper.Viper) (ownerAddress string, ownerType string, er
...
+		// Not set or empty -> print error and stop
+		msg := fmt.Sprintf(
+			"missing workflow owner: when using --%s you must set %q in your config",
+			Flags.RawTxFlag.Name, key,
+		)
+		fmt.Fprintln(os.Stderr, msg)
+		return "", "", errors.New(msg)
+	}
+
+	// if owner is set in settings, use it, higher priority than private key
```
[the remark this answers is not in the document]
shileiwill: not always. `cre secrets execute` doesnt have unsigned flag, but it is a MSIG operation. happy to hear your thoughts on navigating this case.

internal/settings/settings_get.go
```diff
@@ -79,10 +81,39 @@ func GetWorkflowOwner(v *viper.Viper) (ownerAddress string, ownerType string, er
...
+		if owner != "" {
+			if !ethcommon.IsHexAddress(owner) {
+				return "", "", fmt.Errorf("invalid owner address in settings %q: %q", WorkflowOwnerSettingName, owner)
+			}
+			// canonical (checksummed)
+			return ethcommon.HexToAddress(owner).Hex(), constants.WorkflowOwnerTypeMSIG, nil
+		}
 	}
 
+	// TODO I dont think the owner flag is enabled anywhere, remove?
```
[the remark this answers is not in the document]
shileiwill: i can remove this check in the next iteration.

internal/settings/settings_get.go
shileiwill: so if unsigned flag is true, owner-address is ensured to be present? I am not sure that is the case.

test/cli_test.go
```diff
@@ -71,9 +71,11 @@ func createCliEnvFile(envPath string, ethPrivateKey string) error {
 	if err != nil {
 		return err
 	}
-	_, err = writer.WriteString(fmt.Sprintf("%s=%s", settings.EthPrivateKeyEnvVar, ethPrivateKey))
-	if err != nil {
-		return err
+	if strings.TrimSpace(ethPrivateKey) != "" {
```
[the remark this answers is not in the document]
shileiwill: nope, func TrimSpace(s string) string

internal/settings/settings_load.go
```diff
@@ -66,7 +66,7 @@ func AddTxnTypeFlags(cmd *cobra.Command) {
 }
 
 func AddRawTxFlag(cmd *cobra.Command) {
-	cmd.Flags().Bool(Flags.RawTxFlag.Name, false, "If set, the command will return the raw transaction instead of sending it to the network")
+	cmd.Flags().Bool(Flags.RawTxFlag.Name, false, "If set, the command will either return the raw transaction instead of sending it to the network or execute the second step of secrets operations using a previously generated raw transaction")
```
timothyF95: This is a shared flag so we can't mention secrets specifically

timothyF95 (changes requested):

internal/settings/settings_load.go
shileiwill: ack, this is a shared flag, that is why it is an **OR** to make sure we state the complete purpose of the flag.

internal/settings/settings_get.go
```diff
@@ -79,25 +81,50 @@ func GetWorkflowOwner(v *viper.Viper) (ownerAddress string, ownerType string, er
...
+	// unsigned is not set, it is EOA path
 	rawPrivKey := v.GetString(EthPrivateKeyEnvVar)
 	normPrivKey := NormalizeHexKey(rawPrivKey)
 	ownerAddress, err = ethkeys.DeriveEthAddressFromPrivateKey(normPrivKey)
 	if err != nil {
 		return "", "", err
 	}
 
+	// if owner is also set in settings, owner and private key should match
+	if v.IsSet(ownerKey) {
```
shileiwill: Per @timothyF95 :

> I feel quite strongly that this shouldn't be checked. A user will use either an EOA or they will print the transaction to submit themselves with their MSIG. Will they ever have an MSIG private key? also, if they are specifying their EOA, what need is there to also duplicate this with their public key. This will lead to users having to manually switch their configuration every time instead of just using the unsigned flag

I think it is better to have stronger check so users know what is going on, other than using assumptions. Right now, users can have private key in .env and also have owner-address in setting, we dont do any check. we just use the private key under the scene but users are more likely to check the settings and think they use the owner they set there as the address for onchain transactions. Owner address is more explicit than the private key, so i think it is best to check they are aligned.

timothyF95 (approved):

PR #102, opened 2025-10-20
reduce simulate info logs

timothyF95:


cmd/workflow/simulate/telemetry_writer.go
```diff
@@ -432,6 +432,68 @@ func (w *telemetryWriter) mapWorkflowStatus(status string) string {
...
+
+func formatSimulationInfo(message string, fields ...interface{}) {
+	formatSimulationLog(LogLevelInfo, message, fields...)
+}
+
+func formatSimulationWarning(message string, fields ...interface{}) {
+	formatSimulationLog(LogLevelWarning, message, fields...)
+}
+
+func formatSimulationError(message string, fields ...interface{}) {
```
shileiwill: nit: why not just use formatSimulationLog and let the callers decide the log level as in param, it saves some code, and easier for callers

cmd/workflow/simulate/telemetry_writer.go
```diff
@@ -432,6 +432,68 @@ func (w *telemetryWriter) mapWorkflowStatus(status string) string {
 	}
 }
 
+// LogLevel represents different log levels for simulation logs
+type LogLevel string
```
shileiwill: if we apply this to other cli modules, those lines should be moved out of simulation. I think cmd/common is a good place for it.

cmd/workflow/simulate/simulate.go
```diff
@@ -246,18 +246,8 @@ func run(
...
-	logCfg := logger.Config{Level: getLevel(verbosity, zapcore.InfoLevel)}
-	baseLggr, err := logCfg.New()
-	if err != nil {
-		return fmt.Errorf("failed to create logger: %w", err)
-	}
-
+	// Create a minimal logger that discards output since we handle all logging through custom formatters
 	engineLogCfg := logger.Config{Level: zapcore.FatalLevel}
-
-	if inputs.EngineLogs {
```
shileiwill: not sure on this, how will we show engine logs after this change? easy to test with the `-engine-logs` flag on.

engine logs are needed to debug standalone engine. i dont think we need to format these logs. it is for pro users.

cmd/workflow/simulate/telemetry_writer.go
```diff
@@ -13,18 +13,6 @@ import (
 	pb "github.com/smartcontractkit/chainlink-protos/workflows/go/events"
 )
 
-// Color constants for consistent styling
-const (
-	COLOR_RESET       = "\033[0m"
-	COLOR_RED         = "\033[91m" // Bright Red
-	COLOR_GREEN       = "\033[32m" // Green
-	COLOR_YELLOW      = "\033[33m" // Yellow
```
timothyF95: These are not cross platform compatible

ibrajer (approved):

PR #122, opened 2025-10-26
🔔 Warn Users When CLI Version Is Outdated

ejacquier:
### What?

Display a warning when the user is not using the latest CLI version.

image

- Use the GitHub [JSON API](https://api.github.com/repos/smartcontractkit/cre-cli/releases/latest) to check for the latest release and compare it with the current version.  
- Check for a new update every 24 hours and cache the result in `~/.cre`.  
- Display a warning to the user for all CLI commands.  

### Important

To make this work, a few global changes were introduced:

- Added a `sync.WaitGroup` to the update check command to prevent the goroutine from terminating prematurely when the CLI exits quickly (e.g., with commands such as `cre version`).  
- Added a `WaitGroup` for the telemetry command to ensure it completes execution.  
- Added a `Run` function to `rootCmd`, allowing `cre` (without arguments) to display update warnings — not just `cre --help`.  

### Testing

To build the CLI with a custom version, run:  
```bash
go build -ldflags="-X 'github.com/smartcontractkit/cre-cli/cmd/version.Version=v0.1.0'" -o cre main.go
```

You can use the environment variable CRE_FORCE_UPDATE_CHECK to disable caching and force an update check:

```bash
export CRE_FORCE_UPDATE_CHECK=1
```

### TODO
- Add unit tests

### Known Issues
- The update check doesn't display when user pass the --help option. This is a default behavior of Cobra that is harder to override.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -18,6 +18,7 @@ import (
 	"github.com/smartcontractkit/cre-cli/cmd/login"
 	"github.com/smartcontractkit/cre-cli/cmd/logout"
 	"github.com/smartcontractkit/cre-cli/cmd/secrets"
+	"github.com/smartcontractkit/cre-cli/cmd/update"
 	"github.com/smartcontractkit/cre-cli/cmd/version"
 	"github.com/smartcontractkit/cre-cli/cmd/whoami"
 	"github.com/smartcontractkit/cre-cli/cmd/workflow"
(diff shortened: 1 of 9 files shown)


cmd/root.go
```diff
@@ -36,14 +37,24 @@ var runtimeContextForTelemetry *runtime.Context
...
 	err := RootCmd.Execute()
 
 	if err != nil && executingCommand != nil && runtimeContextForTelemetry != nil {
-		telemetry.EmitCommandEvent(executingCommand, 1, runtimeContextForTelemetry)
+
+		telemetryWG.Add(1)
+		go func() {
+			defer telemetryWG.Done()
+			telemetry.EmitCommandEvent(executingCommand, 1, runtimeContextForTelemetry)
+		}()
```
ejacquier: @timothyF95 Please have a look. We'll have to fully test the telemetry after this.
I suspect telemetry would not run for some command like `cre version`

timothyF95 (dismissed):

PR #130, opened 2025-10-28
Init command & template generation improvements

ejacquier:
## What

Update to improve `cre init` UX
- Suggest default for project name and workflow name prompt
- Ask for RPC url (PoR workflow)
- Updated next steps for both go and ts templates
- Updated config so that its different per environment (workflow name and config path)


internal/settings/template/workflow.yaml.tpl
```diff
@@ -1,3 +1,4 @@
+# template/workflow.yaml.tpl
```
anirudhwarrier: @ejacquier is this needed?

PR #145, opened 2025-10-30
Update TS templates to use the latest SDK

ernest-nowacki:
- suggested `cre-sdk` for TS templates updated from `0.0.8-alpha` to `0.0.10-alpha` 
- update `tsconfig` that we generate for new project to rather suggest modern `ESM` over `commonjs`
- updates POR example readme to cover ABIs in TypeScript world


cmd/creinit/template/workflow/typescriptPorExampleDev/README.md
```diff
@@ -54,12 +55,27 @@ Note: The following 7 chains are supported in local simulation (both testnet and
...
+For each contract you would like to interact with, you need to provide the ABI `.ts` file so that TypeScript can provide type safety and autocomplete for the contract methods. The format of the ABI files is very similar to regular JSON format; you just need to export it as a variable and mark it `as const`. For example:
+
+```ts
+// IERC20.ts file
+export const IERC20Abi = {
+  // ... your ABI here ...
+} as const;
+```
+
+For a quick start, every contract used in this workflow is already provided in the `contracts` folder. You can use them as a reference.
```
shileiwill: 👍

cmd/creinit/template/workflow/typescriptSimpleExample/tsconfig.json.tpl
```diff
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
     "target": "esnext",
-    "module": "commonjs",
```
jeffrifwaldsmartcontract: I remember making a mental note to fix this and I guess I never did. Thanks!

shileiwill (approved):
LG!

PR #155, opened 2025-11-03
Point TS SDK to v1.0.0

ernest-nowacki:
Following latest request `cre-sdk` has been re-released with `1.0.0` version (no code changes compared to `1.0.0-beta.1`).

Tested: I did a local build, created two workflows with `cre init` and both worked during simulation.


shileiwill (approved):
LG

PR #157, opened 2025-11-03
Allow patches and minor

ernest-nowacki:
The caret (^) tells bun to install the latest compatible minor or patch version under the same major version of a dependency. I believe in 1.0.0 this should be the way to go.

We wouldn't need to create release of cre cli every time we have a non-breaking patch to ts sdk.

### DIFF
--- a/cmd/creinit/template/workflow/typescriptSimpleExample/package.json.tpl
+++ b/cmd/creinit/template/workflow/typescriptSimpleExample/package.json.tpl
@@ -8,7 +8,7 @@
   },
   "license": "UNLICENSED",
   "dependencies": {
-    "@chainlink/cre-sdk": "1.0.0"
+    "@chainlink/cre-sdk": "^1.0.0"
   },
   "devDependencies": {
     "@types/bun": "1.2.21"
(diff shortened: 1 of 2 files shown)


bradford-m (approved):
we trust an @chainlink package as an internal dependency so we can always allow its latest version. Caret should not be applied to any external dependencies

cmd/creinit/template/workflow/typescriptSimpleExample/package.json.tpl
```diff
@@ -8,7 +8,7 @@
   },
   "license": "UNLICENSED",
   "dependencies": {
-    "@chainlink/cre-sdk": "1.0.0"
+    "@chainlink/cre-sdk": "^1.0.0"
```
shileiwill: to confirm, this will use the latest under the same major version, v1 here. But if we have v2 sdk, this will not auto bump to v2, is that correct?

nolag: It's the desired behaviour, since v2.0.0 could break your workflow :)

PR #159, opened 2025-11-03
feat: sets go sdk to v1

MStreet3:
Sets the cre-sdk-go to v1.0.0 and leaves capabilities at v1.0.0-beta.0


nolag (approved):
If you feel like taking my suggestion, it would be a nice thing to do, but I won't block the PR.

PR #170, opened 2025-11-07
check tool and give explicit error if required tools are not installed

shileiwill:


timothyF95 (approved):
LG overall, one small query

PR #177, opened 2025-11-13
add artifact path check

timothyF95:
```
➜  my-project cre workflow simulate tf-wf-1
Error: failed to load settings: ConfigPath does not exist: dummyFile.txt (configured in your settings for target 'staging-settings')
```

### DIFF
--- a/internal/settings/settings_test.go
+++ b/internal/settings/settings_test.go
@@ -324,3 +324,41 @@ func TestShouldSkipGetOwner(t *testing.T) {
 		})
 	}
 }
+
+func TestArtifactPathValidation(t *testing.T) {
+	envVars := map[string]string{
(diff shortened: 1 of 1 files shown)


internal/settings/workflow_settings.go
```diff
@@ -84,6 +85,11 @@ func loadWorkflowSettings(logger *zerolog.Logger, v *viper.Viper, cmd *cobra.Com
 		return WorkflowSettings{}, errors.Wrap(err, "for target "+target)
 	}
 
+	// Validate artifact paths
+	if err := validateArtifactPaths(&workflowSettings, target); err != nil {
```
j-nowak: What about a test when some of the files is missing? Is it covered already?

PR #194, opened 2025-12-03
align the workflow name 10 bytes logic btw simulator and don

shileiwill:
the standalone engine has the check on the workflow name length, https://github.com/smartcontractkit/chainlink/blob/develop/core/services/workflows/cmd/cre/utils/standalone_engine.go#L85

we had the hashing and truncating in simulator to ensure users can give any length of workflow name, but this makes the 10bytes workflow name doesnt align.  Will investigate why we need the length check in workflow_meta.go

----------
update:
confirmed workflow name length is not a problem in simulator, the standalone engine checks only if the name is empty or too long (256 chars).


shileiwill:
Thanks @miro-cl for confirming via transaction https://sepolia.basescan.org/tx/0xf2e70e3ca03ecbc7e6eea6b0b575424e29cc55a2090356db060f21a9544b1d6e#eventlog

PR #204, opened 2025-12-12
update test to cover simulation on templates

shileiwill:


.github/workflows/pull-request-main.yml
```diff
@@ -14,7 +14,7 @@ env:
 
 jobs:
   ci-lint:
-    runs-on: ubuntu-latest-4cores-16GB
```
shileiwill: the old one is not available any more, we got CI issue, hence the change.

timothyF95 (dismissed):

PR #220, opened 2026-01-09
Add zksync testnet and mainnet to simulator

amit-momin:


amit-momin:
Requires mock forwarders

PR #277, opened 2026-02-20
Add bindings EVM bindings for TS

nolag:
Creates the EVM bindings for TS and uses them in the templates.

For examples of the templates see the files [here](https://github.com/smartcontractkit/cre-cli/tree/bc21013c2fef428543868453029d11ca1b8f4b6d/cmd/creinit/template/workflow/typescriptPorExampleDev/contracts/evm/ts/generated). Since they were removed from the CLI, it's no longer in this PR.


nolag:
> New structure for abi folders make sense now. Good to go, this will need to be updated from main with the new dynamic template. PoR examples now lives in cre-templates and not in the cli.

Yeah, I agreed to merge that after you approve, otherwise this PR doesn't have any context on what the generated output looks like because none of the templates would have any.

PR #279, opened 2026-02-20
Dynamic template fetching from GitHub repos

ejacquier:
# Dynamic template fetching from GitHub repos  

## Summary
- Templates are no longer embedded in the CLI. They are standalone CRE projects hosted in GitHub repos, discovered via template.yaml metadata, downloaded as tarballs, and extracted into the user's workspace. No CLI release needed to add or update templates.                      
- Two built-in Hello World templates (Go + TypeScript) remain embedded as offline fallbacks.                                                                                                                                                           
- New cre templates list|add|remove commands to manage template repository sources, stored in ~/.cre/template.yaml.                                                                                                                                    
- template.yaml standard defines template metadata (kind, id, title, language, category, tags, capabilities, networks, workflows, projectDir, postInit, exclude).
- Templates with projectDir set are treated as complete CRE projects — extracted as-is, skipping config generation and workflow renaming. Go templates with projectDir run go mod tidy post-scaffold.
- Template list (1hr TTL) and tarballs (24hr TTL) are cached at ~/.cre/template-cache/. cre init --refresh bypasses cache.
- Interactive Bubble Tea wizard for template selection with search, language filtering, and network RPC URL prompts.
- CLI prompt helpers (ui.Confirm, ui.Input, ui.Select, ui.InputForm) consolidated from raw huh calls across all commands.

### DIFF
--- a/cmd/generate-bindings/bindings/abigen/FORK_METADATA.md
+++ b/cmd/generate-bindings/bindings/abigen/FORK_METADATA.md
@@ -5,7 +5,7 @@
 - Source Repository: https://github.com/ethereum/go-ethereum
 - Original Package: accounts/abi/bind
 - Fork Date: 2025-06-18
-- Upstream Version: v1.16.0
+- Upstream Version: v1.17.0
 - Upstream Commit: 4997a248ab4acdb40383f1e1a5d3813a634370a6
 
 ## Modifications
@@ -30,6 +30,7 @@
 ## Sync History
 
 - 2025-06-18: Initial fork from v1.16.0
+- 2026-02-25: Bump metadata version (no changes)
 
 ## Security Patches Applied
 


internal/templaterepo/cache.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("failed to write cache: %w", err)
+	}
+
+	c.logger.Debug().Msgf("Saved template list cache for %s", source)
+	return nil
+}
+
+// TarballPath returns the path where a tarball should be cached.
+func (c *Cache) TarballPath(source RepoSource, sha string) string {
+	return filepath.Join(c.cacheDir, "tarballs", fmt.Sprintf("%s-%s-%s.tar.gz", source.Owner, source.Repo, sha))
```
tarcisiozf: we should always sanitize external values when navigation on local paths, to avoid dir traversal and escaping the cache dir

cmd/whoami/whoami.go
```diff
@@ -101,6 +107,15 @@ func (h *Handler) Execute(ctx context.Context) error {
...
 	}
 
+	// Add deployment access status
+	if deployAccess != nil {
+		if deployAccess.HasAccess {
+			details = fmt.Sprintf("%s\nDeploy Access:     Enabled", details)
+		} else {
+			details = fmt.Sprintf("%s\nDeploy Access:     Not enabled", details)
+		}
+	}
```
tarcisiozf:
```suggestion
	if deployAccess != nil && deployAccess.HasAccess {
		details = fmt.Sprintf("%s\nDeploy Access:     Enabled", details)
	} else {
		details = fmt.Sprintf("%s\nDeploy Access:     Not enabled", details)
	}
```

PR #303, opened 2026-03-09
Skip optional api calls ci

timothyF95:


internal/telemetry/telemetry_test.go
```diff
@@ -57,17 +57,24 @@ func TestCollectCommandInfo(t *testing.T) {
...
 	defer func() {
-		if wasSet {
-			os.Setenv(TelemetryDisabledEnvVar, originalValue)
+		if disabledWasSet {
+			os.Setenv(TelemetryDisabledEnvVar, originalDisabled)
 		} else {
 			os.Unsetenv(TelemetryDisabledEnvVar)
 		}
+		if ciWasSet {
+			os.Setenv("CI", originalCI)
```
ibrajer: Will this env var be set in the GHA workflow before running the tests?

PR #338, opened 2026-04-01
chore: replace cre.chain.link with app.chain.link

jihokoo:
The self-serve portal domain has moved from cre.chain.link to app.chain.link to support multiple BCM products.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -3,7 +3,7 @@
 # It detects the architecture, downloads the correct .exe,
 # and adds it to the user's PATH.
 #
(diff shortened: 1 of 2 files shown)


anirudhwarrier (approved):
confirmed installation works using the new link.

anirudhwarrier (changes requested):
@jihokoo repo requires all commits to be signed/verified before it can be merged.

timothyF95:
Replicated here: https://github.com/smartcontractkit/cre-cli/pull/524

PR #365, opened 2026-04-09
Fix EVM Bindings generator struct prefix bug

silaslenihan:
Previously only `tmplStruct.Name` was being updated. `tmplField.Type` is set earlier in bindStructType as a plain string and is what the template uses for struct fields in sourcecre.go.tpl `({{.Type}})`, not the dynamic bindtype path used for method args.

So you could get:
`type Foo struct { ... }` ← name was stripped
`tems []MyContractFoo` ← field still pointed at the old name
That’s invalid Go and fails to compile.

Now we are sure to rename the old field names with the new names

### DIFF
--- a/cmd/generate-bindings/bindings/abigen/bindv2.go
+++ b/cmd/generate-bindings/bindings/abigen/bindv2.go
@@ -380,7 +380,7 @@ func BindV2(types []string, abis []string, bytecodes []string, pkg string, libs 
 	return string(code), nil
 }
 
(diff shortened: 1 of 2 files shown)


nolag (approved):
I'm assuming that it wasn't possible to generate code that was sometimes valid that would now be broken. If that's not the case, you may need to create an alias to the broken name...

ibrajer (approved):

PR #367, opened 2026-04-11
feat: private workflow registry deployment behind a feature flag

j-nowak:
- Added a preview private-registry deploy path for `cre workflow deploy` behind `--preview-private-registry` (STAGING-only).
- Refactored deploy orchestration into target-specific strategies (`onchain` vs `private`) and routed pre-checks/upsert through a shared interface.
- Implemented private-registry upsert flow via GraphQL client, including request payload mapping and deploy result output.
- Updated artifact handling so uploads happen after pre-deploy checks and are skipped when artifact URLs are already provided.
- Added broad test coverage (unit + integration/multi-command flow) for private-registry target resolution, owner derivation, input mapping, and happy-path deploy behavior.


cmd/workflow/deploy/deploy.go
```diff
@@ -275,93 +282,43 @@ func (h *handler) Execute(ctx context.Context) error {
...
-			return nil
-		}
-	} else {
-		if err := h.ensureOwnerLinkedOrFail(); err != nil {
-			return err
-		}
+// resolveWorkflowOwner returns the effective owner address for workflow ID computation.
+// For private registry deploys, the owner is derived from tenantID and organizationID.
+// For onchain deploys, the configured WorkflowOwner address is used directly.
+func (h *handler) resolveWorkflowOwner(targetWorkflowRegistry registryTarget) (string, error) {
```
j-nowak, line 302: Temporary implementation, that will be resolved with derived workflow owner from GQL in the future.

cmd/workflow/deploy/private_registry_test.go
```diff
@@ -0,0 +1,340 @@
...
+	t.Helper()
+	logger := testutil.NewTestLogger()
+	prev, hadPrev := os.LookupEnv(credentials.CreApiKeyVar)
+	os.Setenv(credentials.CreApiKeyVar, "test-key-for-init")
+	creds, err := credentials.New(logger)
+	if hadPrev {
+		os.Setenv(credentials.CreApiKeyVar, prev)
+	} else {
+		os.Unsetenv(credentials.CreApiKeyVar)
+	}
```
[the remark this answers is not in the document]
j-nowak: It's acceptable, similar approach in other tests as well. It would require bigger refactoring.

cmd/workflow/deploy/registry_deploy_strategy_onchain.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return nil
+}
+
+func (h *handler) workflowExists() error {
+	workflow, err := h.wrc.GetWorkflow(common.HexToAddress(h.inputs.WorkflowOwner), h.inputs.WorkflowName, h.inputs.WorkflowTag)
+	if err != nil {
+		return err
+	}
+	if workflow.WorkflowId == [32]byte(common.Hex2Bytes(h.workflowArtifact.WorkflowID)) {
```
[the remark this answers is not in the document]
j-nowak: This compiles

test/multi_command_flows/workflow_private_registry.go
```diff
@@ -0,0 +1,248 @@
...
+	creConfig := "AccessToken: " + jwt + "\n" +
+		"IDToken: test-id-token\n" +
+		"RefreshToken: test-refresh-token\n" +
+		"ExpiresIn: 3600\n" +
+		"TokenType: Bearer\n"
+
+	require.NoError(t, os.WriteFile(filepath.Join(creDir, "cre.yaml"), []byte(creConfig), 0o600), "failed to write test credentials")
+
+	return homeDir
+}
```
j-nowak, line 42: The private registry test needs to override HOME (and consequently GOPATH) because it uses bearer token (JWT) authentication rather than the API key authentication used by the other happy-path e2e tests.

Why HOME must be overridden:

The other on-chain/EOA deploy tests authenticate by simply setting an environment variable (t.Setenv(credentials.CreApiKeyVar, "test-api")). The CLI reads that env var directly — no files on disk involved, no need to touch HOME.
The private registry flow authenticates using bearer credentials stored in a file at ~/.cre/cre.yaml. The test must fabricate a fake cre.yaml containing a synthetic JWT (with the right org_id, exp, etc.) so the CLI subprocess can read it. To make the subprocess find the fake credentials file (instead of the real user's ~/.cre/cre.yaml, or fail because none exists), HOME is redirected to a t.TempDir() where the fake .cre/cre.yaml lives.
Why GOPATH must be pinned:

When HOME is overridden to a temp directory, Go's default GOPATH becomes $HOME/go — which now points inside the temp dir (e.g. /tmp/.../go). The Go module cache under GOPATH contains read-only files. When the test finishes, t.TempDir() cleanup tries to remove the entire temp directory tree. It cannot delete the read-only module cache files, which causes the cleanup to fail and marks the test as failed even though the actual test logic succeeded. By explicitly setting GOPATH back to the real home's go directory (e.g. /Users/jnowak/go), the module cache stays outside the temp dir and cleanup succeeds cleanly.

In short: the on-chain tests inject auth via a simple env var, but the private registry test injects auth via a file on disk, which forces the HOME override, which in turn necessitates the GOPATH fix to avoid a spurious cleanup failure.

cmd/workflow/deploy/deploy.go
```diff
@@ -117,6 +116,7 @@ func New(runtimeContext *runtime.Context) *cobra.Command {
 	deployCmd.Flags().Bool("no-config", false, "Deploy without a config file")
 	deployCmd.Flags().Bool("default-config", false, "Use the config path from workflow.yaml settings (default behavior)")
 	deployCmd.Flags().Bool(cmdcommon.SkipTypeChecksCLIFlag, false, "Skip TypeScript project typecheck during compilation (passes "+cmdcommon.SkipTypeChecksFlag+" to cre-compile)")
+	deployCmd.Flags().Bool("preview-private-registry", false, "Deploy to the private workflow registry (preview, STAGING only)")
```
timothyF95: nit: As staging is not a publicly available environment, could we change this to something like 'This is an unreleased feature'?

cmd/workflow/deploy/deploy.go
timothyF95: Ah, I was going to note this. Would we add a new resolveOwner endpoint? From what I can see this flow does not support API_KEY yet, which is fine for now but will need to be supported eventually

cmd/workflow/deploy/registry_deploy_strategy.go
```diff
@@ -0,0 +1,78 @@
...
+	// Upsert registers or updates the workflow in the target registry
+	// and displays the result.
+	Upsert() error
+}
+
+type registryTargetType string
+
+const (
+	registryTargetOnchain registryTargetType = "onchain"
+	registryTargetPrivate registryTargetType = "private"
```
timothyF95, line 34: Can we reuse the registry types already defined by tenant manifest? https://github.com/smartcontractkit/cre-cli/blob/2c7e30158e5db58e57805199cf3f4c30d20fb145/internal/tenantctx/tenantctx.go#L121

We could hoist as an enum or const as you suggested in https://github.com/smartcontractkit/cre-cli/pull/371#discussion_r3080086162

timothyF95 (dismissed):
Can address in follow up PR

cmd/workflow/deploy/registry_deploy_strategy.go
j-nowak: This part should be refactored in https://smartcontract-it.atlassian.net/browse/DEVSVCS-4634. I guess the flow control will change and use fields directly from resolved registry.

cmd/workflow/deploy/deploy.go
j-nowak: We discussed offline, we will use derieved owner from API call, part of a follow up work.

timothyF95 (approved):

PR #371, opened 2026-04-14
Add deployment registry selector

timothyF95:
## Summary
Adds an optional deployment-registry field to user-workflow settings that lets workflow commands resolve registry configuration dynamically from context.yaml instead of using hardcoded environments.yaml defaults. When omitted, behaviour is unchanged (backward compatible). Off-chain registries are gated in production and rejected by on-chain-only commands (deploy, pause, activate, delete) with a clear error until off-chain routing is implemented.

## Changes:

- New DeploymentRegistry field on UserWorkflowSettings, read via Viper
- ResolveRegistry helper resolves registry by ID from tenant context, branching on type (on-chain / off-chain), with fallback to EnvironmentSet defaults
- ResolvedRegistry wired into runtime.Context; deploy, pause, activate, delete read from it instead of EnvironmentSet directly
- RequireOnChainRegistry guard rejects off-chain registries early with an actionable error message

### Example workflow.yaml
```
staging-settings:
  user-workflow:
    workflow-name: "my-workflow-staging"
    deployment-registry: "onchain:ethereum-testnet-sepolia"           <--- new (optional)
  workflow-artifacts:
    workflow-path: "."
    config-path: "./config.staging.json"
    secrets-path: ""
```

### Precedence hierarchy
```
deployment-registry set?
├── YES → context.yaml registry entry (env vars ignored)
└── NO  → CRE_CLI_WORKFLOW_REGISTRY_* env var
          └── not set → embedded environments.yaml default
```


internal/testutil/testsettings/test_settings.go
```diff
@@ -1,4 +1,4 @@
-package testutil
+package testsettings
```
timothyF95, line 1: This was moved to avoid an import cycle

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/smartcontractkit/cre-cli/internal/environments"
+	"github.com/smartcontractkit/cre-cli/internal/tenantctx"
+)
+
+// ResolvedRegistry holds the fully resolved registry configuration for a
+// workflow command. It is built from either the static EnvironmentSet defaults
+// or from a tenant context registry entry selected via deployment-registry.
+type ResolvedRegistry struct {
+	ID          string
+	Type        string // "on-chain" or "off-chain"
```
j-nowak: Maybe we could extract the type as cosnts/enums?

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	resolved := &ResolvedRegistry{
+		ID:          reg.ID,
+		Type:        reg.Type,
+		DonFamily:   tenantCtx.DefaultDonFamily,
+		ExplorerURL: envSet.WorkflowRegistryChainExplorerURL,
+	}
+
+	if reg.Address != nil {
```
j-nowak: Does it make sense to resolve a registry without an address?

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+
+	"github.com/smartcontractkit/cre-cli/internal/environments"
+	"github.com/smartcontractkit/cre-cli/internal/tenantctx"
+)
+
+// ResolvedRegistry holds the fully resolved registry configuration for a
+// workflow command. It is built from either the static EnvironmentSet defaults
+// or from a tenant context registry entry selected via deployment-registry.
+type ResolvedRegistry struct {
```
j-nowak: Have you considered how the code would look like if we had 2 different types representing onchain/offchain regsitry? I'm not the biggest fan, where we have a bag of fields unrelated for a given object type.

internal/settings/registry_resolution.go
timothyF95: You're right it doesn't. See comment below

internal/settings/registry_resolution.go
timothyF95: I have split into two types `OffChainRegistry` and `OnChainRegistry`

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,172 @@
...
+// tenantCtx.Registries. On-chain entries must have a non-empty address.
+// Off-chain registries are rejected in production environments.
+func ResolveRegistry(
+	deploymentRegistry string,
+	tenantCtx *tenantctx.EnvironmentContext,
+	envSet *environments.EnvironmentSet,
+) (ResolvedRegistry, error) {
+	if deploymentRegistry == "" {
+		return defaultFromEnvironmentSet(envSet), nil
+	}
```
j-nowak, line 83: Maybe we should add a follow up task to plan full migration and clean up? E.g. introduce extra warning and deprecation message, automatically "fix" user settings without this registry, and after some time clean up this branch, after majority of users will migrate.

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if tenantCtx == nil {
+		return nil, fmt.Errorf("deployment-registry %q is set but user context is not available — run `cre login` and retry", deploymentRegistry)
+	}
```
j-nowak, line 87: What if we started the login flow here, so we make the experience more smooth?

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,172 @@
...
+		}, nil
+	}
+
+	address := ""
+	if reg.Address != nil {
+		address = *reg.Address
+	}
+	if address == "" {
+		return nil, fmt.Errorf("on-chain registry %q has no address in context.yaml", reg.ID)
+	}
```
j-nowak: It could be simplified:
```
if reg.Address == nil || *reg.Address == "" {
	return nil, fmt.Errorf("on-chain registry %q has no address in context.yaml", reg.ID)
}
address := *reg.Address
```

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if reg.ChainSelector != nil {
+		sel, err := strconv.ParseUint(*reg.ChainSelector, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("invalid chain_selector %q for registry %q: %w", *reg.ChainSelector, reg.ID, err)
+		}
+		name, err := GetChainNameByChainSelector(sel)
+		if err != nil {
+			return nil, fmt.Errorf("cannot resolve chain name for selector %d (registry %q): %w", sel, reg.ID, err)
+		}
```
j-nowak: It seems like the whole logic could be part of `GetChainNameByChainSelector` or some equivalent function

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,172 @@
...
+
+	address := ""
+	if reg.Address != nil {
+		address = *reg.Address
+	}
+	if address == "" {
+		return nil, fmt.Errorf("on-chain registry %q has no address in context.yaml", reg.ID)
+	}
+
+	resolved := &OnChainRegistry{
```
j-nowak: Why not resolve chain selector earlier and create the struct fully at once? I would move it after chain selector parsing.

internal/settings/registry_resolution.go
```diff
@@ -0,0 +1,172 @@
...
+
+// ResolvedRegistry is the interface implemented by both OnChainRegistry and
+// OffChainRegistry. Commands type-switch on the concrete type to access
+// type-specific fields, which prevents accidental use of fields that don't
+// exist for a given registry kind.
+type ResolvedRegistry interface {
+	GetID() string
+	GetType() RegistryType
+	GetDonFamily() string
+}
```
j-nowak, line 27: Nit: AI told me this is more idiomatic structure:
```
type ResolvedRegistry interface {
    ID() string
    Type() RegistryType
    DonFamily() string
}

type OnChainRegistry struct {
    id          string
    address     string
    chainName   string
    donFamily   string
    explorerURL string
}

func (r *OnChainRegistry) ID() string { return r.id }
```

cmd/workflow/deploy/deploy.go
```diff
@@ -174,14 +174,19 @@ func (h *handler) ResolveInputs(v *viper.Viper) (Inputs, error) {
...
+		SkipConfirmation:       v.GetBool(settings.Flags.SkipConfirmation.Name),
+		SkipTypeChecks:         v.GetBool(cmdcommon.SkipTypeChecksCLIFlag),
+		PreviewPrivateRegistry: previewPrivateRegistry,
+		TargetWorkflowRegistry: targetWorkflowRegistry,
+	}
+
+	if oc, ok := h.runtimeContext.ResolvedRegistry.(*settings.OnChainRegistry); ok {
+		inputs.WorkflowRegistryContractChainName = oc.ChainName
+		inputs.WorkflowRegistryContractAddress = oc.Address
+	}
```
j-nowak, line 187: Later, as part of the task https://smartcontract-it.atlassian.net/browse/DEVSVCS-4634 we will be able to do something like:
```
switch r := registry.(type) {
case *OnChainRegistry:
    return deployOnChain(r)
case *OffChainRegistry:
    return deployOffChain(r)
}
```

In my opinion looks relatively good.

A better desing, but it might be hard to achieve due to current structure and coupling in the application, we could have behavioral functions on the registry type itself like `Deploy()`, `Activate()`, etc. That way the registry would encapsulate the client and we would have clear aggreagte of interactions with the registry. But in my experience with the `deploy` command, it's not achievable in short time, as handler logic is leaking implementation details among many fields and functions, making it very difficult.

internal/settings/registry_resolution.go
timothyF95: This is a good suggestion 👍

internal/settings/registry_resolution.go
timothyF95: Can you clarify what you mean by automatically "fix"? Fetch tenantCtx again and rebuild struct?

internal/settings/registry_resolution.go
timothyF95: This is true

internal/settings/registry_resolution.go
timothyF95: Hmm.. you're right. This will have a ripple affect of needing to add a constructor function as they are unexported.

cmd/workflow/deploy/deploy.go
timothyF95: I've added this as a note to the ticket for ref

internal/settings/registry_resolution.go
timothyF95: So I looked into this and it's more complicated than I first thought. I suggest we follow up as a UX enhancement and keep this PR scope down

internal/settings/registry_resolution.go
j-nowak: Yes, run the login flow in the background to fetch tenant ctx

internal/settings/settings_get.go
```diff
@@ -250,6 +251,17 @@ func GetChainNameByChainSelector(chainSelector uint64) (string, error) {
 	return chainDetails.ChainName, nil
 }
 
+// ChainNameFromSelectorString parses a raw chain-selector string and resolves
+// it to a chain name. It combines the string-to-uint64 conversion with the
+// selector-to-name lookup in a single call.
+func ChainNameFromSelectorString(raw string) (string, error) {
```
j-nowak, line 257: nit: we could have some tests for that

internal/settings/registry_resolution.go
j-nowak: Can you add a task for it?

j-nowak (approved):
Approved, can you add tickets for follow ups?

PR #372, opened 2026-04-14
[DEVSVCS-4628] Simulator - Use new cron capability

tarcisiozf:
Refactored both interactive and non-interactive cron trigger flows to use a `skipWaitSignal` channel, allowing the trigger to be skipped or controlled more flexibly (e.g., by pressing Enter in interactive mode). This replaces the previous use of `time.Now()` and improves user experience and testability.


cmd/workflow/simulate/simulate.go
```diff
@@ -741,7 +741,26 @@ func makeBeforeStartInteractive(holder *TriggerInfoAndBeforeStart, inputs Inputs
...
+
+				done, err := triggerCaps.ManualCronTrigger.ManualTrigger(cronCtx, triggerRegistrationID)
+				if err != nil {
+					cancel()
+					return err
+				}
+
+				go func() {
+					defer cancel()
+					<-done
```
jmank88: `ManualTrigger` might send a value before closing the channel. What are we meant to be listening for here?

tarcisiozf: added a comment with context

jmank88: The problem is that only waiting for one signal when there may be two is a race condition. What is the intention of this code? Why send on the channel at all?

cmd/workflow/simulate/simulate.go
```diff
@@ -749,7 +749,31 @@ func makeBeforeStartInteractive(holder *TriggerInfoAndBeforeStart, inputs Inputs
...
+				}
+
+				go func() {
+					// After trigger event happens we can stop the user prompt goroutine if it's still waiting for input
+					defer cancel()
+					defer close(skipWaitSignal)
+					<-done
+				}()
+
+				return nil
```
jmank88, line 773: Why is it OK to return here while there are still goroutines in flight?

tarcisiozf: yeah, the actual trigger lifecycle gets updated in a separated channel here: https://github.com/smartcontractkit/chainlink/pull/22012/changes#diff-ff3f5bf4a05f391df1803ae09579cbba8def8c0f530804c2be2d72e37975dc08R182

jmank88: Would it be valid to move the body of the first goroutine here? If not, why not?
```suggestion
				ui.Line()
				pressed := ui.WaitForEnter(userPromptCtx, "Cron scheduler started. Press Enter to skip waiting...")
				if pressed {
					skipWaitSignal <- struct{}{}
				}

				return nil
```

tarcisiozf: that works too

anirudhwarrier (approved):

PR #373, opened 2026-04-15
refactor: make simulator chain-agnostic with pluggable chain families

Fletch153:
## Why

The simulator was EVM-only by construction. Adding a second chain family meant editing dozens of call sites across `simulate.go`.

## What changed

EVM is now one implementation of a chain-family interface instead of the only shape the simulator understands. EVM-specific code — clients, capabilities, health checks, trigger parsing, CLI flags, limits lives under its own package. The simulator iterates registered families rather than calling EVM directly.

No user-facing behaviour changes. No CLI, config, or limit changes.

## Out of scope / follow-ups

- **Second chain family (Aptos / Solana).** The whole point of this refactor. Will surface any remaining EVM assumptions the interface didn't flush out. This PR is focused purely on relocating code and making subtle improvements to the modularity.


cawthorne:
It looks good from my perspective.

Can we add any extra integration test which tests all these new pieces together, rather than just unit tests?

cmd/workflow/simulate/chain/evm/trigger.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+
+	evmpb "github.com/smartcontractkit/chainlink-common/pkg/capabilities/v2/chain-capabilities/evm"
+	valuespb "github.com/smartcontractkit/chainlink-protos/cre/go/values/pb"
+
+	"github.com/smartcontractkit/cre-cli/internal/ui"
+)
+
+// parse "ChainSelector:<digits>" from trigger id, e.g. "evm:ChainSelector:5009297550715157269@1.0.0 LogTrigger"
```
amit-momin: Is this format really an EVM specific thing? The parser below seems pretty agnostic. It would work for Aptos too if the cap ID would just be `aptos:ChainSelector:4741433654826277614@1.0.0` for mainnet as an example.

cmd/workflow/simulate/simulate.go
```diff
@@ -744,58 +702,59 @@ func makeBeforeStartInteractive(holder *TriggerInfoAndBeforeStart, inputs Inputs
...
+				if !ok {
+					continue
+				}
 
-			client := inputs.EVMClients[sel]
-			if client == nil {
-				ui.Error(fmt.Sprintf("No RPC configured for chain selector %d", sel))
-				os.Exit(1)
-			}
+				if !ct.HasSelector(sel) {
```
amit-momin: Wouldn't we want to continue in case the next chain type has the selector?

Fletch153: Technically yes but that wasn't how it was intended. The logic should have been:

- Loop each family
- Parse the family/selector combination
- If found, check it's been registered

Instead, the second step was only checking the selector, so if EVM was looping first it would've just parsed the selector out of an Aptos trigger ID, not found it registered, and failed before Aptos ever got a turn. It should be fixed now.

cmd/workflow/simulate/chain/evm/trigger.go
Fletch153: Moved this into the package up as i can't see anytime soon having family specific parsing

amit-momin (dismissed):

tarcisiozf (dismissed):

tarcisiozf (dismissed):

tarcisiozf (approved):

PR #375, opened 2026-04-15
Fix: --limits none now fully disables engine-level limit enforcement   

ejacquier:
# Summary                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                  
  --limits none was intended to disable all limit enforcement during simulation, but only suppressed CLI-side checks (WASM binary size, HTTP request/response     
  size). The engine itself still applied its built-in defaults from cresettings.Default — for example HTTPAction.CallLimit = 5 — because ResolveLimits("none")
  returns nil and the WorkflowSettingsCfgFn callback skipped applyEngineLimits when simLimits == nil, leaving the engine to fall back to its own defaults.        
                  
  This caused unexpected errors like cannot use 6, limit is 5 even when --limits none was explicitly passed.                                                      
   
  ## Fix                                                                                                                                                             
                  
  cmd/workflow/simulate/limits.go — adds disableEngineLimits(cfg *cresettings.Workflows), which sets all engine limit fields to effectively unbounded values      
  (math.MaxInt32 for int/size fields, 24h for duration fields). 

This covers all capability categories: execution, WASM, logging, HTTPAction, ConfidentialHTTP, Consensus, ChainWrite, ChainRead, and Secrets.                                                                                                                  
                  
  cmd/workflow/simulate/simulate.go — in the WorkflowSettingsCfgFn callback, adds an else if inputs.LimitsPath == "none" branch that calls disableEngineLimits(cfg), so the engine receives the uncapped values instead of falling back to its built-in defaults.


timothyF95 (approved):

cmd/workflow/simulate/limits.go
```diff
@@ -93,6 +97,70 @@ func applyEngineLimits(cfg *cresettings.Workflows, limits *SimulationLimits) {
...
+// effectively disabling limit enforcement when --limits none is used.
+func disableEngineLimits(cfg *cresettings.Workflows) {
+	maxInt := settings.Setting[int]{DefaultValue: math.MaxInt32}
+	maxSize := settings.Setting[config.Size]{DefaultValue: math.MaxInt32}
+	maxDuration := settings.Setting[time.Duration]{DefaultValue: 24 * time.Hour}
+
+	// Execution limits
+	cfg.ExecutionTimeout = maxDuration
+	cfg.ExecutionResponseLimit = maxSize
+	cfg.ExecutionConcurrencyLimit = maxInt
```
timothyF95, line 110: nit: These limits seem quite excessive? A buggy workflow could OOM a test machine

PR #379, opened 2026-04-16
Refactor deploy command to be registry agnostic

j-nowak:
Move onchain registry dependency on the onchain deployment flow.
Partially refactor `SimualtedEnvironment` to work with both onchain & offchain registries.

### DIFF
--- a/cmd/workflow/deploy/autoLink.go
+++ b/cmd/workflow/deploy/autoLink.go
@@ -97,14 +97,6 @@ func (h *handler) autoLinkMSIGAndExit(onChain *settings.OnChainRegistry) (halt b
 	return true, nil
 }
 
-func linkKeyInputsForAutoLink(h *handler, onChain *settings.OnChainRegistry) linkkey.Inputs {
-	return linkkey.Inputs{
-		WorkflowOwner:                   h.inputs.WorkflowOwner,
-		WorkflowRegistryContractAddress: onChain.Address(),
-		WorkflowOwnerLabel:              h.inputs.OwnerLabel,
-	}
-}
-
 // tryAutoLink executes the auto-link process using the link-key command
 func (h *handler) tryAutoLink(onChain *settings.OnChainRegistry) error {
 	rtx := &runtime.Context{
@@ -115,7 +107,11 @@ func (h *handler) tryAutoLink(onChain *settings.OnChainRegistry) error {
 		EnvironmentSet: h.environmentSet,
 	}
 
-	return linkkey.Exec(rtx, linkKeyInputsForAutoLink(h, onChain))
+	return linkkey.Exec(rtx, linkkey.Inputs{
+		WorkflowOwner:                   h.inputs.WorkflowOwner,
+		WorkflowRegistryContractAddress: onChain.Address(),
+		WorkflowOwnerLabel:              h.inputs.OwnerLabel,
+	})
 }
 
 // checkLinkStatusViaGraphQL checks if the owner is linked and verified by querying the service
--- a/cmd/workflow/deploy/autoLink.go
+++ b/cmd/workflow/deploy/autoLink.go
@@ -13,6 +13,7 @@ import (
 	linkkey "github.com/smartcontractkit/cre-cli/cmd/account/link_key"
 	"github.com/smartcontractkit/cre-cli/internal/client/graphqlclient"
 	"github.com/smartcontractkit/cre-cli/internal/runtime"
+	"github.com/smartcontractkit/cre-cli/internal/settings"
 	"github.com/smartcontractkit/cre-cli/internal/ui"
 )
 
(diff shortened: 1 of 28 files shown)


internal/testutil/chainsim/simulated_environment.go
```diff
@@ -56,6 +57,20 @@ func (se *SimulatedEnvironment) NewRuntimeContextWithBufferedOutput() (*runtime.
...
+	return ctx
+}
+
+func (se *SimulatedEnvironment) NewOffChainRuntimeContextWithBufferedOutput(tenantID, donFamily string) (*runtime.Context, *bytes.Buffer) {
+	ctx, buf := se.NewRuntimeContextWithBufferedOutput()
+	ctx.TenantContext = &tenantctx.EnvironmentContext{TenantID: tenantID}
+	ctx.ResolvedRegistry = settingspkg.NewOffChainRegistry("private", donFamily)
+	return ctx, buf
+}
+
```
j-nowak, line 73: Temporary solution, that should be refactored later. But those changes are faster and unblock further development.
https://smartcontract-it.atlassian.net/browse/DEVSVCS-4650

cmd/workflow/deploy/autoLink.go
```diff
@@ -13,6 +13,7 @@ import (
 	linkkey "github.com/smartcontractkit/cre-cli/cmd/account/link_key"
 	"github.com/smartcontractkit/cre-cli/internal/client/graphqlclient"
 	"github.com/smartcontractkit/cre-cli/internal/runtime"
+	"github.com/smartcontractkit/cre-cli/internal/settings"
```
ibrajer, line 16: Ultimate nit picker: Can we rename the file to auto_link.go because it really sticks out 🙈

cmd/workflow/deploy/compile_test.go
```diff
@@ -64,10 +62,8 @@ func TestCompileCmd(t *testing.T) {
 			{
 				name: "Non-ASCII ConfigPath",
 				cmd: Inputs{
-					WorkflowPath:                      "testdata/test_workflow.yaml",
-					ConfigPath:                        "./testdata/đuveč.yaml",
-					WorkflowRegistryContractAddress:   "0x1234567890123456789012345678901234567890",
-					WorkflowRegistryContractChainName: "ethereum-testnet-sepolia",
+					WorkflowPath: "testdata/test_workflow.yaml",
+					ConfigPath:   "./testdata/đuveč.yaml",
```
ibrajer, line 66: I can't believe that đuveč survived this long

cmd/workflow/deploy/deploy_test.go
```diff
@@ -539,10 +540,12 @@ func TestValidateInputs_PrivateRegistry(t *testing.T) {
 		)
 		h.settings = ctx.Settings
 		h.environmentSet.EnvName = "STAGING"
-		h.environmentSet.DonFamily = ""
-		h.runtimeContext.ResolvedRegistry = settings.NewOffChainRegistry("private", "")
+		token := makeTestJWT(t, map[string]interface{}{
+			"sub":    "user1",
+			"org_id": "org-test-123",
+		})
```
ibrajer, line 546: Can we bake this test JWT into `simulatedEnvironment`? It's always the same value, right?

cmd/workflow/deploy/registry_deploy_strategy_onchain.go
```diff
@@ -17,12 +18,18 @@ import (
...
 	wrc     *client.WorkflowRegistryV2Client
+	onChain *settings.OnChainRegistry
 	wg      sync.WaitGroup
 	initErr error
 }
 
-func newOnchainRegistryDeployStrategy(h *handler) *onchainRegistryDeployStrategy {
-	a := &onchainRegistryDeployStrategy{h: h}
+func newOnchainRegistryDeployStrategy(h *handler) (*onchainRegistryDeployStrategy, error) {
+	onChain, err := settings.AsOnChain(h.runtimeContext.ResolvedRegistry, "deploy")
```
ibrajer, line 27: Instead of using raw strings, can we create a typed string (enum) for on-chain commands? Or whatever `"deploy"` is meant to represent here.

Another option is the builder pattern
`settings.AsOnChain(h.runtimeContext.ResolvedRegistry).Deploy()`

ibrajer (approved):
Approving to unblock, comments can be addressed in the next PRs, because they are mostly nits.

cmd/workflow/deploy/deploy_test.go
j-nowak: Added to task https://smartcontract-it.atlassian.net/browse/DEVSVCS-4650

cmd/workflow/deploy/autoLink.go
j-nowak: Will do https://smartcontract-it.atlassian.net/browse/DEVSVCS-4659

cmd/workflow/deploy/registry_deploy_strategy_onchain.go
j-nowak: As we roll out to everyone, this will be changed and this function won't survive fully:
```
// AsOnChain asserts that r is an *OnChainRegistry. If it is not, it returns a
// descriptive error mentioning the command that required on-chain support.
func AsOnChain(r ResolvedRegistry, commandName string) (*OnChainRegistry, error) {
	if oc, ok := r.(*OnChainRegistry); ok {
		return oc, nil
	}
	return nil, fmt.Errorf(
		"%s currently only supports on-chain registries; deployment-registry %q is %s",
		commandName, r.ID(), r.Type(),
	)
}
```
This `commandName` is used only for error.

PR #382, opened 2026-04-17
Implement `cre workflow pause` for private registry

j-nowak:


cmd/workflow/pause/pause.go
```diff
@@ -135,158 +107,40 @@ func (h *handler) Execute() error {
...
-		}
-
-		workflows = append(workflows, list...)
-
-		start = big.NewInt(start.Int64() + int64(len(list)))
-		if int64(len(list)) < pageSize {
-			break
-		}
+	if len(owner) >= 2 && owner[:2] != "0x" {
+		owner = "0x" + owner
```
ibrajer, line 134: I don't understand why this is needed - can't we ensure that `DerivedWorkflowOwner` propagated to the runtime context is a valid ETH address instead? It can't even be empty, it always has to exist

ibrajer (approved):

PR #390, opened 2026-04-20
Resolve workflow owner by deployment registry

timothyF95:
When user-workflow.deployment-registry is set, workflow owner is no longer resolved during the first settings load (which required a local key even for off-chain registries). Owner is filled after the registry is resolved: off-chain uses the session-derived owner with type ORG_DERIVED; on-chain still uses GetWorkflowOwner. Derived owner is normalized once when credentials are validated (FormatWorkflowOwnerAddress). Per-command 0x prefix handling in workflow handlers is removed. Tests and workflow fixture YAMLs added/updated.

### DIFF
--- a/internal/ethkeys/keys.go
+++ b/internal/ethkeys/keys.go
@@ -10,20 +10,20 @@ import (
 )
 
 // FormatWorkflowOwnerAddress trims whitespace, ensures a 0x prefix for hex input,
-// and returns the address in EIP-55 checksummed form. Returns "" if s is empty
-// or not a valid 20-byte hex address.
-func FormatWorkflowOwnerAddress(s string) string {
+// and returns the address in EIP-55 checksummed form. Empty input (after trim)
+// returns ("", nil). Non-empty input that is not a valid 20-byte hex address returns an error.
+func FormatWorkflowOwnerAddress(s string) (string, error) {
 	s = strings.TrimSpace(s)
 	if s == "" {
-		return ""
+		return "", nil
 	}
 	if len(s) < 2 || (s[0:2] != "0x" && s[0:2] != "0X") {
 		s = "0x" + s
 	}
 	if !common.IsHexAddress(s) {
-		return ""
+		return "", fmt.Errorf("invalid owner address %q", s)
 	}
-	return common.HexToAddress(s).Hex()
+	return common.HexToAddress(s).Hex(), nil
 }
 
 func DeriveEthAddressFromPrivateKey(privateKeyHex string) (string, error) {
--- a/internal/ethkeys/keys.go
+++ b/internal/ethkeys/keys.go
@@ -3,9 +3,28 @@ package ethkeys
 import (
 	"crypto/ecdsa"
 	"fmt"
+	"strings"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 )
+
+// FormatWorkflowOwnerAddress trims whitespace, ensures a 0x prefix for hex input,
+// and returns the address in EIP-55 checksummed form. Returns "" if s is empty
+// or not a valid 20-byte hex address.
+func FormatWorkflowOwnerAddress(s string) string {
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return ""
+	}
+	if len(s) < 2 || (s[0:2] != "0x" && s[0:2] != "0X") {
+		s = "0x" + s
+	}
+	if !common.IsHexAddress(s) {
+		return ""
+	}
+	return common.HexToAddress(s).Hex()
+}
 
 func DeriveEthAddressFromPrivateKey(privateKeyHex string) (string, error) {
 	privateKey, err := crypto.HexToECDSA(privateKeyHex)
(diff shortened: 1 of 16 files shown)


internal/ethkeys/keys.go
```diff
@@ -3,10 +3,29 @@ package ethkeys
...
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return ""
+	}
+	if len(s) < 2 || (s[0:2] != "0x" && s[0:2] != "0X") {
+		s = "0x" + s
+	}
+	if !common.IsHexAddress(s) {
+		return ""
+	}
```
j-nowak, line 25: Shouldn't this be an error?

internal/settings/workflow_settings.go
```diff
@@ -163,19 +170,61 @@ func loadWorkflowSettings(logger *zerolog.Logger, v *viper.Viper, cmd *cobra.Com
...
+		}
+		ownerType = constants.WorkflowOwnerTypeOrgDerived
+	} else {
+		ownerAddr, ownerType, err = GetWorkflowOwner(v)
+		if err != nil {
+			return err
+		}
+	}
+	workflow.UserWorkflowSettings.WorkflowOwnerAddress = ownerAddr
+	workflow.UserWorkflowSettings.WorkflowOwnerType = ownerType
```
j-nowak, line 215: Have you considered not making this part of settings at all? What if we added a dynamic `runtimeContext.ResolveOwner(<params like registry and whatever else we might need>)`?

internal/settings/workflow_settings.go
```diff
@@ -118,18 +120,23 @@ func loadWorkflowSettings(logger *zerolog.Logger, v *viper.Viper, cmd *cobra.Com
...
+	workflowSettings.UserWorkflowSettings.DeploymentRegistry = getSetting(DeploymentRegistrySettingName)
+	deploymentRegistry := workflowSettings.UserWorkflowSettings.DeploymentRegistry
+
+	// If deployment-registry is set, owner depends on how that id resolves; defer to
+	// FinalizeWorkflowOwner (after ResolveRegistry). Otherwise resolve from env/config now.
 	if !ShouldSkipGetOwner(cmd) {
-		ownerAddress, ownerType, err := GetWorkflowOwner(v)
-		if err != nil {
-			return WorkflowSettings{}, err
+		if deploymentRegistry == "" {
```
j-nowak, line 129: Can `deploymentRegistry` still be empty here?

internal/settings/workflow_settings.go
```diff
@@ -163,19 +170,61 @@ func loadWorkflowSettings(logger *zerolog.Logger, v *viper.Viper, cmd *cobra.Com
...
 	return workflowSettings, nil
 }
 
+// FinalizeWorkflowOwner sets workflow owner when loadWorkflowSettings deferred it because
+// user-workflow.deployment-registry was non-empty. Call after ResolveRegistry.
+func FinalizeWorkflowOwner(
+	v *viper.Viper,
+	cmd *cobra.Command,
+	workflow *WorkflowSettings,
+	target string,
```
j-nowak, line 186: What is target?

internal/settings/workflow_settings.go
```diff
@@ -163,19 +170,61 @@ func loadWorkflowSettings(logger *zerolog.Logger, v *viper.Viper, cmd *cobra.Com
...
+) error {
+	if ShouldSkipGetOwner(cmd) {
+		return nil
+	}
+	if workflow.UserWorkflowSettings.DeploymentRegistry == "" {
+		return nil
+	}
+	if resolved == nil {
+		return fmt.Errorf("resolved registry is required to finalize workflow owner")
+	}
```
j-nowak, line 198: Do we need to check for both?

internal/ethkeys/keys.go
ibrajer: Same question for L18

internal/settings/workflow_settings.go
timothyF95: This is closer to to what we will need as a final solution. I think we are going to have to break the settings down into smaller components and load on request. Will share details offline

internal/settings/workflow_settings.go
timothyF95: Target is your settings target/group e.g. `staging-settings`

internal/settings/workflow_settings.go
timothyF95: Yes, to maintain backwards compatibility, we must account for an empty deploymentRegistry else this would be a breaking change to current projects. If it is not set, getSetting returns an empty string

internal/ethkeys/keys.go
timothyF95: This is a good point, it should error

internal/settings/workflow_settings.go
timothyF95: Short answer is yes. If DeploymentRegistry is "" then we should early return as this step is not needed. If resolved is nil then settings are in a broken state and cannot proceed.

internal/settings/workflow_settings.go
j-nowak: Ok, I thought we load registry before making this call?

j-nowak (approved):

PR #397, opened 2026-04-22
Add workflow list command

timothyF95:


cmd/workflow/list/list.go
```diff
@@ -0,0 +1,428 @@
...
+			return NewHandler(runtimeContext).Execute(cmd.Context(), registryID, includeDeleted)
+		},
+	}
+
+	cmd.Flags().StringVar(&registryID, "registry", "", "Filter by registry ID from user context")
+	cmd.Flags().BoolVar(&includeDeleted, "include-deleted", false, "Include workflows in DELETED status")
+	return cmd
+}
+
+func (h *Handler) fetchAllWorkflows(ctx context.Context) ([]workflowRow, error) {
```
j-nowak: This whole functions seems like should rather be part of a specific workflow data client.

cmd/workflow/list/list.go
```diff
@@ -0,0 +1,428 @@
...
+
+		batch := envelope.Workflows.Data
+		all = append(all, batch...)
+
+		if len(all) >= total || len(batch) == 0 {
+			break
+		}
+	}
+
+	return all, nil
```
j-nowak: It's better to return internal model rather than couple the implementation to type returned by external client.

cmd/workflow/list/list.go
```diff
@@ -0,0 +1,428 @@
...
+	if reg == nil || !hasContractAddress(reg) {
+		return false
+	}
+	if registryTypeOffChain(reg) {
+		return false
+	}
+	return true
+}
+
+// registryEligibleForGrpcRows is true for off-chain registries and for legacy entries
```
j-nowak: What are legacy entries? I don't understand.

cmd/workflow/list/list.go
```diff
@@ -0,0 +1,428 @@
...
+	out := make([]workflowRow, 0, len(rows))
+	for _, r := range rows {
+		if rowMatchesRegistry(r.WorkflowSource, reg, all) {
+			out = append(out, r)
+		}
+	}
+	return out
+}
+
+func rowMatchesRegistry(workflowSource string, reg *tenantctx.Registry, all []*tenantctx.Registry) bool {
```
j-nowak: Can't we just comapre workflowSource with full form of a `reg` as it's stored in manifest? Do we have a mismatch there?

cmd/workflow/list/list_test.go
```diff
@@ -0,0 +1,810 @@
...
+	f.call++
+	var body []byte
+	var err error
+
+	switch f.call {
+	case 1:
+		body, err = json.Marshal(map[string]any{
+			"workflows": map[string]any{
+				"count": 3,
+				"data": []map[string]string{
```
j-nowak: Maybe workflow source should have real forms?

cmd/workflow/list/list_test.go
j-nowak, line 1: It would be much easier to test, if we had smaller modules instead all-in-one handler:
* client module with GQL dependency
* ui module just printing workflow table
* handler module calling client to fetch workflows, applying some transformation/filtering and then calling ui to render them

Right now to test a small thing/error we need to properly configure all dependencies.

cmd/workflow/list/list.go
timothyF95: Verbiage is AI slop but it caters for unknown registry types

cmd/workflow/list/list.go
timothyF95: Will break it down into modules

cmd/workflow/list/list.go
timothyF95: Will add a workflow data type

cmd/workflow/list/list.go
timothyF95: Yes we have a miss match. the way it is returned from the API is different to how we have it in the manifest which is why we map on registry address

cmd/workflow/list/list_client.go
```diff
@@ -0,0 +1,100 @@
...
+      status
+      workflowSource
+    }
+    count
+  }
+}
+`
+
+// Executor runs a GraphQL request (e.g. graphqlclient.Client).
+type Executor interface {
```
j-nowak: I think more natural name would be something like `WorkflowDataClient`. Also, it could be externalized simialr as privateregistryclient or storageclient. In the future we might want to replace it with CRE SDK, to avoid writing gql queries manually.

cmd/workflow/list/list.go
```diff
@@ -0,0 +1,99 @@
...
+// NewHandler builds a handler with the real GraphQL executor.
+func NewHandler(ctx *runtime.Context) *Handler {
+	return NewHandlerWithClient(ctx, nil)
+}
+
+// NewHandlerWithClient builds a handler with an optional GraphQL executor (nil uses graphqlclient.New).
+func NewHandlerWithClient(ctx *runtime.Context, gql Executor) *Handler {
+	if gql == nil {
+		gql = graphqlclient.New(ctx.Credentials, ctx.EnvironmentSet, ctx.Logger)
+	}
```
j-nowak: Both storage and privateregistry clients have gql as their dependency. I think it would be good to have the same consistent way to structure such clients.

cmd/workflow/list/list_client_test.go
j-nowak, line 1: What about using `httptest.NewServer` for those client tests? We use it in other clients.

cmd/workflow/list/list_test.go
```diff
@@ -0,0 +1,809 @@
...
+	if err := cmd.Execute(); err == nil {
+		t.Fatal("expected error when extra args provided")
+	}
+}
+
+type fakeGQL struct {
+	call int
+}
+
+func (f *fakeGQL) Execute(ctx context.Context, req *graphql.Request, resp any) error {
```
j-nowak: Why not use `httptest.NewServer`? It's closer to the real execution flow

cmd/workflow/list/registry.go
```diff
@@ -0,0 +1,214 @@
...
+	out := make([]Workflow, 0, len(rows))
+	for _, r := range rows {
+		if rowMatchesRegistry(r.WorkflowSource, reg, all) {
+			out = append(out, r)
+		}
+	}
+	return out
+}
+
+func rowMatchesRegistry(workflowSource string, reg *tenantctx.Registry, all []*tenantctx.Registry) bool {
```
j-nowak, line 27: This function deserves independent tests

cmd/workflow/list/registry_test.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+func sp(s string) *string { return &s }
+
+func TestRowMatchesRegistry_DirectIDMatch(t *testing.T) {
+	reg := &tenantctx.Registry{ID: "private", Type: "off-chain"}
+	all := []*tenantctx.Registry{reg}
+
+	if !rowMatchesRegistry("private", reg, all) {
+		t.Error("expected direct ID match to return true")
```
j-nowak, line 16: Why not use `assert.True` or `assert.False`? It's more clear than `if` and `t.Error`.

cmd/workflow/list/registry.go
j-nowak, line 27: Right now  looking also at tests, the name of this function could be better. E.g. `workflowSourcesMatchesRegistry`. WDYT?

j-nowak (approved):

PR #398, opened 2026-04-23
Add list output flag

timothyF95:


cmd/workflow/list/list.go
```diff
@@ -18,6 +20,36 @@ import (
 // can use the name without importing workflowdataclient directly.
 type Workflow = workflowdataclient.Workflow
 
+// Inputs holds the resolved and validated flag values for the list command.
+type Inputs struct {
+	RegistryFilter string
+	IncludeDeleted bool
+	// OutputPath is an absolute path to a .json file, or "" if not set.
+	OutputPath string
```
j-nowak: Do we want an `OutputPath` or just `JSON` format? In terminal you can always redirect the output to the file if you want with `>>`. If the purpose for it is to be used by AI agents, I think it would be more convenient to have JSON output printed to stdout, where they can scrape it easily.

j-nowak (approved):

PR #402, opened 2026-04-27
Simulator empty log message check

ejacquier:
# Summary

Display an error message `empty message payload not allowed; use "\n" within a message for formatting` when workflow has an empty `runtime.log()`.

### DIFF
--- a/cmd/workflow/simulate/telemetry_writer.go
+++ b/cmd/workflow/simulate/telemetry_writer.go
@@ -253,6 +253,16 @@ func (w *telemetryWriter) handleCapabilityEvent(telLog TelemetryLog, eventType s
 	}
 }
 
+// trailingNewlines returns the number of consecutive "\n" characters at the
+// end of s.
+func trailingNewlines(s string) int {
+	n := 0
+	for i := len(s) - 1; i >= 0 && s[i] == '\n'; i-- {
+		n++
+	}
+	return n
+}
+
 // formatUserLogs formats and displays user logs in a readable way
 func (w *telemetryWriter) formatUserLogs(logs *pb.UserLogs) {
 	// Display each log line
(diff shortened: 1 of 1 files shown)


cmd/workflow/simulate/telemetry_writer.go
```diff
@@ -256,6 +257,29 @@ func (w *telemetryWriter) handleCapabilityEvent(telLog TelemetryLog, eventType s
...
+				StyleRed.Render(`empty message payload not allowed; use "\n" within a message for formatting`),
+				StyleRed,
+			)
+			continue
+		}
+
+		// Whitespace-only payloads (e.g. runtime.log("\n")) render as actual
+		// blank line(s) so users can add visual spacing between log sections.
+		if strings.TrimSpace(logLine.Message) == "" {
+			n := strings.Count(logLine.Message, "\n")
```
ibrajer: If you have a string "this is \n my error \n details \n more \n" it will add 4 new lines, instead of one.
So you need to count the number of "\n" at the end of the string

cmd/workflow/simulate/telemetry_writer.go
```diff
@@ -252,10 +253,47 @@ func (w *telemetryWriter) handleCapabilityEvent(telLog TelemetryLog, eventType s
...
+			continue
+		}
+
+		// Whitespace-only payloads (e.g. runtime.log("\n")) render as actual
+		// blank line(s) so users can add visual spacing between log sections.
+		// Only count *trailing* newlines: counting every "\n" in the message
+		// would over-emit blank lines for content like
+		// "this is \n my error \n details \n more \n" (4 newlines, but the
+		// user only intends 1 blank line of trailing spacing).
+		if strings.TrimSpace(logLine.Message) == "" {
```
timothyF95, line 286: This means that if a user adds `logLine.Message = " " ` (with a 1 or more spaces) it will log 1 blank line. Is that what we want?

timothyF95:
Discussed async - we will not be making this change in the CLI

PR #410, opened 2026-04-30
Fix rpc and eth key validation bug

timothyF95:


internal/settings/workflow_settings.go
```diff
@@ -319,6 +321,11 @@ func ShouldSkipGetOwner(cmd *cobra.Command) bool {
 		b, _ := cmd.Flags().GetBool("broadcast")
 		return !b
 	default:
+		// Browser OAuth flow resolves the owner from credentials, not from
+		// CRE_ETH_PRIVATE_KEY, so skip EOA owner derivation at settings load.
+		if f := cmd.Flags().Lookup("secrets-auth"); f != nil && f.Value.String() == "browser" {
```
anirudhwarrier, line 326: looks like adding this to skip list has some unintended consequences.
- build from this branch - `secrets list --secrets-auth=browser` shows only org owned secrets
- build from main - `secrets list --secrets-auth=browser` shows both legacy (owner owned) and org owned secret.

anirudhwarrier (changes requested):

internal/settings/workflow_settings.go
anirudhwarrier: confirmed by @timothyF95 that this is expected behavior.

anirudhwarrier (approved):

PR #412, opened 2026-05-04
fix: secrets oauth skip getOwner

anirudhwarrier:
[DEVSVCS-4808](https://smartcontract-it.atlassian.net/browse/DEVSVCS-4808)

**Refined logic for skipping owner retrieval:**

* Updated `ShouldSkipGetOwner` to check full command paths and handle additional cases: now skips for `cre workflow hash`, for `cre workflow simulate` when `--broadcast` is false or unset, and for any `cre secrets ...` command when `--secrets-auth` is set to `browser`.
* Added new test cases to `TestShouldSkipGetOwner` for various `cre secrets` command scenarios, specifically covering the new logic for skipping when `--secrets-auth` is `browser`.

### DIFF
--- a/internal/settings/workflow_settings.go
+++ b/internal/settings/workflow_settings.go
@@ -305,20 +305,23 @@ func IsValidChainName(name string) error {
 }
 
 // For commands that don't need the private key, we skip getting the owner address.
-// ShouldSkipGetOwner returns true if the command is `simulate` and
-// `--broadcast` is false or not set. `cre help` should skip as well.
+// ShouldSkipGetOwner returns true for `cre workflow hash`; for `cre workflow simulate` when
+// `--broadcast` is false or unset; and for any `cre secrets ...` command when `--secrets-auth` is browser.
 func ShouldSkipGetOwner(cmd *cobra.Command) bool {
-	switch cmd.Name() {
-	case "help":
+	path := cmd.CommandPath()
+	switch path {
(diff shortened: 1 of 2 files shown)


internal/settings/workflow_settings.go
```diff
@@ -305,20 +305,23 @@ func IsValidChainName(name string) error {
...
 
 // For commands that don't need the private key, we skip getting the owner address.
-// ShouldSkipGetOwner returns true if the command is `simulate` and
-// `--broadcast` is false or not set. `cre help` should skip as well.
+// ShouldSkipGetOwner returns true for `cre workflow hash`; for `cre workflow simulate` when
+// `--broadcast` is false or unset; and for any `cre secrets ...` command when `--secrets-auth` is browser.
 func ShouldSkipGetOwner(cmd *cobra.Command) bool {
-	switch cmd.Name() {
-	case "help":
+	path := cmd.CommandPath()
```
anirudhwarrier, line 311: `cre help` is listed under isLoadSettings’s exclusions in [cmd/root.go](https://github.com/smartcontractkit/cre-cli/blob/931827c102a289fe23b315cc02a35caabba1d9c1/cmd/root.go#L483-L484), so that path does not call AttachSettings / loadWorkflowSettings at all. In practice ShouldSkipGetOwner never runs for normal cre help.

PR #415, opened 2026-05-06
feat: add UI warning for EnvironmentSet values

anirudhwarrier:
[DEVSVCS-4904](https://smartcontract-it.atlassian.net/browse/DEVSVCS-4904)

This pull request enhances environment configuration handling in `internal/environments/environments.go` by improving how environment variables are loaded and introducing clearer feedback for users when environment settings are overridden or missing. It also ensures warnings are only shown once per process.

Key improvements:

**Environment variable handling and user feedback:**

* Refactored the way environment variables are loaded into the `EnvironmentSet`, reducing code duplication and making the logic clearer. Now, each variable is explicitly loaded and checked before assignment.
* Added a mechanism to warn users (via `ui.Warning`) when specific environment variables are set, or when an unknown environment is requested. This helps users understand which configuration is active and if any overrides are in effect. Warnings are shown only once per process using `sync.Once`.

### DIFF
--- a/internal/environments/environments.go
+++ b/internal/environments/environments.go
@@ -137,12 +137,9 @@ func NewEnvironmentSet(ff *fileFormat, envName string) *EnvironmentSet {
 
 	newEnvironmentSetWarningsOnce.Do(func() {
 		switch envName {
-		case DevelopmentEnv:
-			ui.Warning(fmt.Sprintf("%s set, using %s environment", EnvVarEnv, DevelopmentEnv))
-		case StagingEnv:
-			ui.Warning(fmt.Sprintf("%s set, using %s environment", EnvVarEnv, StagingEnv))
 		case DefaultEnv:
-			break
+		case DevelopmentEnv, StagingEnv:
+			ui.Warning(fmt.Sprintf("%s set, using %s environment", EnvVarEnv, envName))
 		default:
 			ui.Warning(fmt.Sprintf("Environment %s not found, defaulting to %s", envName, DefaultEnv))
 		}
--- a/internal/environments/environments.go
+++ b/internal/environments/environments.go
@@ -5,8 +5,11 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"gopkg.in/yaml.v2"
+
+	"github.com/smartcontractkit/cre-cli/internal/ui"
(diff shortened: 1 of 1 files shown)


internal/environments/environments.go
```diff
@@ -78,46 +83,100 @@ func loadEmbeddedEnvironmentFile() (*fileFormat, error) {
...
+		case DevelopmentEnv:
+			ui.Warning(fmt.Sprintf("%s set, using %s environment", EnvVarEnv, DevelopmentEnv))
+		case StagingEnv:
+			ui.Warning(fmt.Sprintf("%s set, using %s environment", EnvVarEnv, StagingEnv))
+		case DefaultEnv:
+			break
+		default:
+			ui.Warning(fmt.Sprintf("Environment %s not found, defaulting to %s", envName, DefaultEnv))
+		}
+		if authBase != "" {
```
timothyF95, line 146: Could we add these warnings to the same checks above so we're not duplicating?

internal/environments/environments.go
```diff
@@ -78,46 +83,100 @@ func loadEmbeddedEnvironmentFile() (*fileFormat, error) {
...
+		set.DonFamily = donFamily
+	}
+	if secretsOrgOwned != "" {
+		set.SecretsOrgOwned = strings.EqualFold(secretsOrgOwned, "true")
+	}
+
+	newEnvironmentSetWarningsOnce.Do(func() {
+		switch envName {
+		case DevelopmentEnv:
+			ui.Warning(fmt.Sprintf("%s set, using %s environment", EnvVarEnv, DevelopmentEnv))
```
timothyF95: If these cases are the same could we not add an if not production -> warn ?

internal/environments/environments.go
anirudhwarrier: The checks are now duplicated to ensure the warning messages are displayed only once per CLI session. This is because `environments.New()` is invoked multiple times during the lifecycle of a CLI session

internal/environments/environments.go
anirudhwarrier: simplified the switch case a bit. keeping switch case as it is easier to read and can expand to more envs easily.

ejacquier (approved):

PR #418, opened 2026-05-07
Document deployment-registry usage in cre registry list

ejacquier:
# Summary                                                                                                                                                                                                                                             
  - cre registry list --help now explains that the displayed ID is the value to set as deployment-registry in workflow.yaml, with a yaml snippet.
  - The command output appends a "Targeting a registry" hint pointing users to set deployment-registry to one of the listed IDs.


anirudhwarrier (approved):
LGTM

PR #424, opened 2026-05-12
fix: align workflow hash owner for private registry

j-nowak:
## Summary
- load credentials for workflow hash to support private registries
- use derived org owner for private/off-chain hash computation
- add unit coverage for hash owner resolution

## Test plan
- go test ./cmd/workflow/hash
- go test ./cmd/...


anirudhwarrier (changes requested):
`workflow hash` should work without login

PR #425, opened 2026-05-12
Update secrets onchain flag name

timothyF95:


cmd/secrets/common/validate.go
```diff
@@ -6,23 +6,23 @@ import (
...
 
 // ValidateSecretsAuthFlow checks that the chosen auth flow is valid and
 // allowed in the current environment. Browser flow is blocked in production.
 func ValidateSecretsAuthFlow(flow, envName string) error {
 	switch flow {
-	case SecretsAuthOwnerKeySigning:
+	case SecretsAuthOnchain:
 		return nil
 	case SecretsAuthBrowser:
 		if strings.EqualFold(envName, "PRODUCTION") || envName == "" {
```
j-nowak, line 20: Are w case insensitive here?

j-nowak (approved):

cmd/secrets/common/validate.go
timothyF95: Yes

PR #426, opened 2026-05-13
Add don family support for offchain registry

timothyF95:


internal/settings/registry_resolution_test.go
j-nowak, line 1: I think using `assert.NoError(t, err)` and `assert.Equal(t, expected, actual)` is easier to read in tests

anirudhwarrier (approved):

PR #429, opened 2026-05-13
Fix installer not getting the version number

frederickjansen:
GitHub minifies their JSON output so a `grep` match returns the whole string, resulting in the installer trying to download version `mentions_name`. Using `jq` fixes this.


timothyF95:
Thank you @frederickjansen, the team will investigate further.

PR #440, opened 2026-05-19
Implement SecureRemove function for safer file deletion and update logout handler to use it.

anirudhwarrier:
[DEVSVCS-4953](https://smartcontract-it.atlassian.net/browse/DEVSVCS-4953)

Enhance the security of credential file deletion by ensuring sensitive data is securely erased before removal. The main change is the introduction of a new `SecureRemove` function, which overwrites credential files with zeroes before deleting them. The logout process and associated tests have also been updated to use this secure deletion method.

**Security improvements:**

* Added a new `SecureRemove` function in `internal/credentials/credentials.go` that overwrites files with zeroes before deleting them, preventing potential recovery of sensitive data.
* Updated the logout handler in `cmd/logout/logout.go` to use `credentials.SecureRemove` instead of `os.Remove` for deleting the credentials file.

### DIFF
--- a/cmd/logout/logout.go
+++ b/cmd/logout/logout.go
@@ -88,7 +88,7 @@ func (h *handler) execute() error {
 		}
 	}
 
-	if err := os.Remove(credPath); err != nil && !os.IsNotExist(err) {
+	if err := credentials.SecureRemove(credPath); err != nil {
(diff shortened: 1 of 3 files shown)


timothyF95 (approved):
LGTM

PR #459, opened 2026-05-28
Fix context propagation in execute operations

timothyF95:
### JIRA
https://smartcontract-it.atlassian.net/browse/DEVSVCS-4957


j-nowak (dismissed):

cmd/workflow/deploy/deploy.go
```diff
@@ -133,6 +133,15 @@ func newHandler(ctx *runtime.Context, stdin io.Reader) *handler {
...
 
+// executionContext returns the context from Execute(), or context.Background()
+// when handler methods are invoked directly in unit tests.
+func (h *handler) executionContext() context.Context {
+	if h.execCtx != nil {
+		return h.execCtx
+	}
+	return context.Background()
+}
+
```
j-nowak: I don't like the pattern, where in the production code we have a branch for testing. It is a code smell and creates a higher risk for bugs and issues. 

Why can't we have one of those:
1. Create handler with `context` as constructor param
2. Pass `ctx` as a function params where needed and then don't store it as a `handler` field?

cmd/workflow/deploy/deploy.go
j-nowak: This is a code smell, and can cause a lot of weird bugs in the future. We shouldn't mutate a field like that.

j-nowak (changes requested):

cmd/workflow/deploy/deploy.go
timothyF95: I think it would be better to pass the ctx as a parameter. It did make testing challenging

j-nowak (approved):

PR #463, opened 2026-06-02
Extract internal/creconfig for CLI config paths

timothyF95:
Summary
- Add `internal/creconfig` as the single owner of CLI config directory paths
- Migrate credentials, `tenantctx`, `templateconfig`, `templaterepo/cache`, `update`, `telemetry`, and related `cmd/tests`
- Remove duplicated `.cre` path helpers from credentials
- Use OS-correct paths in runtime feedback (FilePathHint); keep relative paths in static `help/docs`


cmd/logout/logout_test.go
```diff
@@ -17,8 +18,8 @@ import (
 
 func setupCredentialFile(t *testing.T, home string, token string) {
 	t.Helper()
-	dir := filepath.Join(home, credentials.ConfigDir)
-	if err := os.MkdirAll(dir, 0o700); err != nil {
+	dir, err := creconfig.EnsureDir()
```
j-nowak, line 21: Can we name the variable and function a bit more explicit? It looks quite generic now.

docs/cre_templates_add.md
```diff
@@ -4,7 +4,7 @@ Adds a template repository source
 
 ### Synopsis
 
-Adds one or more template repository sources to ~/.cre/template.yaml. These repositories are used by cre init to discover available templates.
+Adds one or more template repository sources to .cre/template.yaml. These repositories are used by cre init to discover available templates.
```
j-nowak: Is it intentional to remove information that this file is stored by default in the hoemdir? Maybe we could write something like `$HOME_DIR/.cre/template.yaml`

internal/creconfig/creconfig.go
```diff
@@ -0,0 +1,63 @@
+package creconfig
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const Dir = ".cre"
```
j-nowak, line 9: I think it could be beneficial to use more explicit name. Or is it against Go rules?

internal/creconfig/creconfig.go
```diff
@@ -0,0 +1,63 @@
...
+	dir, err := DirPath()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(dir, name), nil
+}
+
+// FileRelPath returns a home-relative path for static help text and committed docs.
+// It uses the host OS path separator (e.g. ".cre/template.yaml" or ".cre\template.yaml").
+func FileRelPath(name string) string {
```
j-nowak: Maybe we should name it `FileHomeRelPath`?

internal/creconfig/creconfig.go
```diff
@@ -0,0 +1,63 @@
...
+	return filepath.Join(Dir, name)
+}
+
+// FilePathHint returns the absolute config file path for user-facing messages,
+// or a home-relative path if the home directory cannot be resolved.
```
j-nowak: How does it work if we can't resolve home dir?

internal/creconfig/creconfig.go
timothyF95: This is not named well. It's specifically used for generating docs/help because it keeps the output clean of os specific syntax. I'm going to remove it as there are only two impacted spaces

internal/creconfig/creconfig.go
timothyF95: In the context of the package I think this is clear enough. Evocation would look like `creconfig.Dir` and it is a go convention to use short names in small single purpose packages

docs/cre_templates_add.md
timothyF95: It is intentional and it's to keep the information OS agnostic. I don't think we should add a $HOME literal but I will add a note in the desc that it is in your home dir

cmd/logout/logout_test.go
timothyF95: IMO this does encapsulate precisely what it is. We have created a new package with specific ownership over your cre configuration creconfig  which is in your home directory EnsureDir makes sure the directory exists. I did consider calling it something similar to creglobalconfig but we may want to extend this in the future to be split between global and local in a similar way that git differentiates between the two. Do you have any ideas of how we could make it more explicit?

internal/creconfig/creconfig.go
timothyF95: it will return `.cre/<filename.ext>`

cmd/logout/logout_test.go
j-nowak: Maybe we could name the `dir` -> `creConfigDir`?

internal/creconfig/creconfig.go
j-nowak: Ok

j-nowak (approved):

PR #471, opened 2026-06-04
Add RPC resolution and consent for capability registry

timothyF95:


cmd/secrets/common/vault_validation.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+const vaultValidationSkippedWarning = "Vault gateway validation skipped; the encryption key and response signatures will not be verified independently of the gateway."
+
+// EnsureVaultValidationOrConsent resolves CapabilitiesRegistry RPC settings and either
+// enables on-chain validation (skipValidation=false) or obtains explicit consent to
+// proceed without validation. The result is cached for the lifetime of the Handler so
+// encrypt and response parsing in the same command only prompt once.
+func (h *Handler) EnsureVaultValidationOrConsent(ctx context.Context) (skipValidation bool, err error) {
+	_ = ctx
```
tarcisiozf:
```suggestion
func (h *Handler) EnsureVaultValidationOrConsent(_ context.Context) (skipValidation bool, err error) {
```

internal/settings/capabilities_registry_rpc.go
```diff
@@ -0,0 +1,48 @@
...
+	expectedSelector := tenantCtx.CapabilitiesRegistry.ChainSelector
+
+	chainName, err = GetChainNameByChainSelector(expectedSelector)
+	if err != nil {
+		return "", "", false, fmt.Errorf("capabilities registry chain selector %d: %w", expectedSelector, err)
+	}
+
+	rpcURL, err = GetRpcUrlSettings(v, chainName)
+	if err != nil {
+		if strings.Contains(err.Error(), "rpc url not found") {
```
tarcisiozf: string matching on errors is not a good practice, can we move this to a sentinel error?

timothyF95: We don't currently use sentinel errors but I think a better solution to string lookup would be to wrap the get in a lookup function as in this case it isn't truly an error

tarcisiozf (approved):

PR #477, opened 2026-06-09
Enable real simulation of Triggers using Local Simulator (with fake capabilities)

tarcisiozf:
https://smartcontract-it.atlassian.net/browse/DEVSVCS-4432


cmd/workflow/simulate/capabilities.go
```diff
@@ -18,10 +18,14 @@ import (
 	"github.com/smartcontractkit/chainlink/v2/core/capabilities/fakes"
 )
 
+// httpTriggerServerPort is the port on which the local HTTP server listens
+// when no --http-payload flag is supplied and the user chooses to POST the payload.
+const httpTriggerServerPort = 2000
```
timothyF95, line 23: nit: With the addition of --listen flag we should probably have a dynamic port here to avoid conflicts. Can be follow up.

cmd/workflow/simulate/simulate.go
```diff
@@ -103,7 +110,8 @@ func New(runtimeContext *runtime.Context) *cobra.Command {
 	simulateCmd.MarkFlagsMutuallyExclusive("config", "no-config", "default-config")
 	// Non-interactive trigger selection flags
 	simulateCmd.Flags().Int("trigger-index", -1, "Index of the trigger to run (0-based)")
-	simulateCmd.Flags().String("http-payload", "", "HTTP trigger payload as JSON string or path to JSON file (with or without @ prefix)")
+	simulateCmd.Flags().String("http-payload", "", "HTTP trigger payload as JSON string or path to JSON file")
+	simulateCmd.Flags().Bool("listen", false, "Listen for HTTP requests or supported log triggers and run the simulator for each match")
```
timothyF95, line 114: Might be worth mentioning in docs that cron is not supported as this may be unclear

timothyF95 (approved):

PR #480, opened 2026-06-10
Add CapabilitiesRegistry client and shared vault DON resolution for secrets flows

timothyF95:
## Summary
- Adds a read-only CapabilitiesRegistry client under `internal/onchain/capabilitiesregistry/` for tenant-scoped on-chain reads via validated project RPC
- Introduces `cmd/secrets/common/vaultdon/` to resolve the vault DON, public key, and OCR signer metadata from the registry
- Wires resolver construction into the secrets handler when CapabilitiesRegistry RPC validation is enabled
- Organizes on-chain contract clients under `internal/onchain/` to prepare for consolidating other registry clients in a follow-up change https://smartcontract-it.atlassian.net/browse/DEVSVCS-5221


cmd/secrets/common/vaultdon/resolver.go
```diff
@@ -0,0 +1,167 @@
...
+// VaultPublicKeyHex returns the on-chain vault master public key for the resolved DON.
+func VaultPublicKeyHex(v *VaultDON) (string, error) {
+	if v == nil {
+		return "", fmt.Errorf("vault DON is nil")
+	}
+	raw, err := vaultCapabilityConfigBytes(v)
+	if err != nil {
+		return "", err
+	}
+	cfg, err := capabilitiesregistry.ParseVaultCapabilityConfig(raw)
```
cedric-cordenier: Nit: small style point -- but you've made the capabilitiesregistry lib aware of how to parse the vault configuration; a cleaner separation would be to just have it parse capability configurations and then have your vaultdon resolver understand what the expected format is

cmd/secrets/common/vaultdon/resolver.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return "", err
+	}
+	return cfg.VaultPublicKey, nil
+}
+
+// OCRAllowedSigners returns the OCR signer addresses for vault DON member nodes.
+func OCRAllowedSigners(v *VaultDON) ([]common.Address, error) {
```
cedric-cordenier: Similar nit: all dons will have signers in this format so arguably this should be moved to capabilitiesregistry

cmd/secrets/common/vaultdon/resolver.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, err
+	}
+
+	r.cached = &VaultDON{DON: vaultDONs[0], Nodes: nodes}
+	return r.cached, nil
+}
+
+// VaultPublicKeyHex returns the on-chain vault master public key for the resolved DON.
+func VaultPublicKeyHex(v *VaultDON) (string, error) {
+	if v == nil {
```
cedric-cordenier, line 94: We should be calling this if VaultDON is nil, same comment with OCRAllowedSigners and MinOCRSignatures

The last one is the most serious, presumably you are using this to understand how many signatures you need to validate. If the function returns 0 on a nil input; this might sometimes return 0 signatures required, which would be bad.

timothyF95: You're right this is a good catch. It should definitely error.

internal/onchain/capabilitiesregistry/nodes.go
```diff
@@ -0,0 +1,26 @@
...
+
+// OCRSignerAddresses returns the OCR signer addresses for CapabilitiesRegistry member nodes.
+func OCRSignerAddresses(nodes []capreg.INodeInfoProviderNodeInfo) ([]common.Address, error) {
+	signers := make([]common.Address, 0, len(nodes))
+	for _, node := range nodes {
+		if len(node.Signer) < 20 {
+			return nil, fmt.Errorf("node signer address too short for p2p id %x", node.P2pId)
+		}
+		signers = append(signers, common.BytesToAddress(node.Signer[:20]))
+	}
```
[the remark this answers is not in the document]
timothyF95: Copilot is wong here

internal/onchain/capabilitiesregistry/nodes_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	signers, err := OCRSignerAddresses([]capreg.INodeInfoProviderNodeInfo{
+		{P2pId: [32]byte{1}, Signer: signerA},
+		{P2pId: [32]byte{2}, Signer: signerB},
+	})
+	require.NoError(t, err)
+	require.Equal(t, []common.Address{
+		common.BytesToAddress(signerA[:20]),
+		common.BytesToAddress(signerB[:20]),
+	}, signers)
```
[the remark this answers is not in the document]
timothyF95: Also here

cmd/secrets/common/vaultdon/resolver_test.go
```diff
@@ -0,0 +1,183 @@
...
+	pubKey, err := vaultdon.VaultPublicKeyHex(v)
+	require.NoError(t, err)
+	require.Equal(t, "deadbeef", pubKey)
+
+	signers, err := capabilitiesregistry.OCRSignerAddresses(v.Nodes)
+	require.NoError(t, err)
+	require.Equal(t, []common.Address{
+		common.BytesToAddress(signerA[:20]),
+		common.BytesToAddress(signerB[:20]),
+	}, signers)
```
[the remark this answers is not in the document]
timothyF95: Also wrong here

internal/onchain/capabilitiesregistry/nodes.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+
+	capreg "github.com/smartcontractkit/chainlink-evm/gethwrappers/workflow/generated/capabilities_registry_wrapper_v2"
+)
+
+// OCRSignerAddresses returns the OCR signer addresses for CapabilitiesRegistry member nodes.
+// Signer is a bytes32 on-chain; the address is stored in the first 20 bytes (see chainlink launcher).
+func OCRSignerAddresses(nodes []capreg.INodeInfoProviderNodeInfo) ([]common.Address, error) {
```
anirudhwarrier: error is always nil.

cmd/secrets/common/vaultdon/resolver_test.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (m *mockReader) GetDONsInFamily(_ context.Context, _ string) ([]*big.Int, error) {
+	return m.donIDs, nil
+}
+
+func (m *mockReader) GetDON(_ context.Context, donID uint32) (capreg.CapabilitiesRegistryDONInfo, error) {
+	don, ok := m.dons[donID]
+	if !ok {
+		return capreg.CapabilitiesRegistryDONInfo{}, context.Canceled
```
anirudhwarrier: If the donID is not found then it should be an error returned, not cancelled context

internal/onchain/capabilitiesregistry/client.go
```diff
@@ -0,0 +1,96 @@
...
+// validated rpcURL format and chain ID against the tenant chain selector.
+func NewReadOnlyClient(ctx context.Context, rpcURL, contractAddress string) (*Client, error) {
+	if !common.IsHexAddress(contractAddress) {
+		return nil, fmt.Errorf("invalid capabilities registry address %q", contractAddress)
+	}
+
+	backend, err := ethclient.DialContext(ctx, rpcURL)
+	if err != nil {
+		return nil, fmt.Errorf("dial capabilities registry RPC: %w", err)
+	}
```
anirudhwarrier, line 36: ethClient is not closed properly. Consider doing defer ethclient.Close()

anirudhwarrier (changes requested):

anirudhwarrier (dismissed):

anirudhwarrier (approved):

PR #481, opened 2026-06-10
Fix browser secrets failure when default CRE_ETH_PRIVATE_KEY placeholder is set

timothyF95:
### JIRA
https://smartcontract-it.atlassian.net/browse/DEVSVCS-5190
### Summary
- Treat the cre-init placeholder (your-eth-private-key) and empty values as unset so browser-auth secrets commands no longer fail on decode
- Add ResolveEthPrivateKeyFromEnv to centralize validation at settings load; invalid non-placeholder values return a clear hex-length error
- Introduce EthPrivateKeyHex with redacted String()/GoString() and explicit .Hex() accessor to reduce accidental key leakage
- Update secrets handler and EVM simulate to use .IsSet()/.Hex() instead of raw string checks
- Share placeholder constant between .env generation and resolution logic

### DIFF
--- a/internal/settings/private_key_env.go
+++ b/internal/settings/private_key_env.go
@@ -1,0 +1,45 @@
+package settings
+
+import (
+	"strings"
+
+	"github.com/smartcontractkit/cre-cli/internal/ethkeys"
+)
+
+// privateKeyEnvPlaceholders are default template values from cre init that should
+// be treated as unset rather than validated as hex private keys.
+var privateKeyEnvPlaceholders = map[string]struct{}{
+	DefaultEthPrivateKeyEnvPlaceholder: {},
+}
+
+// ResolveEthPrivateKeyFromEnv interprets CRE_ETH_PRIVATE_KEY for settings load.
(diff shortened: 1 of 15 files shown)


internal/settings/private_key_env.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	norm := NormalizeHexKey(raw)
+	if _, err := ethkeys.DeriveEthAddressFromPrivateKey(norm); err != nil {
+		return "", err
+	}
+	return EthPrivateKeyHex(norm), nil
+}
+
+// IsUsablePrivateKeyHex reports whether raw looks like an intentional private key
```
anirudhwarrier: This is not used anywhere other than test. Is this planned to be used elsewhere?

anirudhwarrier (changes requested):

internal/settings/private_key_env.go
timothyF95: Ah yes, good catch. This was left over from a convenience helper that was abandoned. Removed

anirudhwarrier (approved):

PR #490, opened 2026-06-12
Validate vault encryption key using CapabilitiesRegistry 

timothyF95:
## Summary
- Verify encryption key against CapabilitiesRegistry when RPC is available: EncryptSecrets still fetches the vault master TDH2 public key from the gateway (publicKey/get). When CapabilitiesRegistry RPC is configured in project.yaml, the CLI also reads the on-chain key from the tenant registry (vault@1.0.0 on the resolved vault DON) and fails if they do not match. With no RPC configured, encryption continues using the gateway key only.
- Remove the temporary vaultValidationGateEnabled bypass and all related test skips so vault validation tests run unconditionally.
- Update unit and integration tests: mock gateway + CapabilitiesRegistry resolver for encrypt/compare paths, deploy a minimal cap reg on Anvil for the secrets happy path (RPC + compare), restore publicKey/get in the gateway mock, and fix GraphQL test mocks to use the correct anvil-devnet chain selector.


test/contracts/vault_cap_reg.go
```diff
@@ -0,0 +1,92 @@
...
+	vaultcommon "github.com/smartcontractkit/chainlink-common/pkg/capabilities/actions/vault"
+	capabilitiespb "github.com/smartcontractkit/chainlink-common/pkg/capabilities/pb"
+	capreg "github.com/smartcontractkit/chainlink-evm/gethwrappers/workflow/generated/capabilities_registry_wrapper_v2"
+	"github.com/smartcontractkit/chainlink-protos/cre/go/values"
+	"github.com/smartcontractkit/chainlink-testing-framework/seth"
+)
+
+// DeployVaultCapabilitiesRegistry deploys a minimal CapabilitiesRegistry with one vault DON
+// for CLI secrets integration tests.
+func DeployVaultCapabilitiesRegistry(t *testing.T, sethClient *seth.Client, vaultPublicKeyHex, donFamily string) common.Address {
```
anirudhwarrier, line 19: @timothyF95 this could be pre-deployed and baked into the anvil state that the tests use. This would reduce test runtime. This is already done for WorkflowRegistry deployment.

anirudhwarrier (approved):
LGTM. Tests optimisation can be a follow-up.

test/contracts/vault_cap_reg.go
timothyF95: Cleated a follow up: https://smartcontract-it.atlassian.net/browse/DEVSVCS-5357

PR #497, opened 2026-06-18
Validate OCR response payload signatures

timothyF95:
### JIRA
https://smartcontract-it.atlassian.net/browse/DEVSVCS-4955


cmd/secrets/common/vault_response_verify.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil
+	}
+
+	signed := rpcResp.Result
+	if signed == nil {
+		return fmt.Errorf("empty SignedOCRResponse result")
+	}
+	if len(signed.Signatures) == 0 {
+		return nil
+	}
```
anirudhwarrier, line 40: @timothyF95 if no signatures are there should the verification go thorugh?

cmd/secrets/common/vault_response_verify.go
```diff
@@ -0,0 +1,59 @@
...
+		return fmt.Errorf("vault response signature verification failed: %w", err)
+	}
+	return nil
+}
+
+func (h *Handler) vaultResponseVerifyContext() context.Context {
+	if h.execCtx != nil {
+		return h.execCtx
+	}
+	return context.Background()
```
anirudhwarrier: when is execution context nil - is it only for tests?

timothyF95: Yea, that's not good. Removed that so that tests instantiate their own context

anirudhwarrier (approved):
LGTM

PR #499, opened 2026-06-22
Update HTTP listening documentation

tarcisiozf:
This pull request adds support for configuring the HTTP trigger server port in the workflow simulator via a new CLI flag, rather than relying on a hardcoded value. The port can now be set with the `--http-trigger-port` flag, and this value is propagated throughout the codebase, including tests and documentation. The default port remains 2000 if not specified.

**HTTP Trigger Port Configuration:**

* Introduced a new CLI flag `--http-trigger-port` (default: 2000) to allow users to specify the port for the local HTTP trigger server in `simulate.go`, and updated the `Inputs` struct and input resolution logic to support this parameter.

### DIFF
--- a/docs/cre_workflow_simulate.md
+++ b/docs/cre_workflow_simulate.md
@@ -28,8 +28,9 @@ cre workflow simulate ./my-workflow
       --evm-tx-hash string           EVM trigger transaction hash (0x...)
   -h, --help                         help for simulate
       --http-payload string          HTTP trigger payload as JSON string or path to JSON file
+      --http-trigger-port int        Port used by the local HTTP trigger server (default 2000)
       --limits string                Production limits to enforce during simulation: 'default' for prod defaults, path to a limits JSON file (e.g. from 'cre workflow limits export'), or 'none' to disable (default "default")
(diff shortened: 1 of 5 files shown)


docs/cre_workflow_simulate.md
```diff
@@ -28,8 +28,9 @@ cre workflow simulate ./my-workflow
       --evm-tx-hash string           EVM trigger transaction hash (0x...)
   -h, --help                         help for simulate
       --http-payload string          HTTP trigger payload as JSON string or path to JSON file
+      --http-trigger-port int        Port used by the local HTTP trigger server (default 2000)
       --limits string                Production limits to enforce during simulation: 'default' for prod defaults, path to a limits JSON file (e.g. from 'cre workflow limits export'), or 'none' to disable (default "default")
-      --listen                       Listen for HTTP requests or supported log triggers and run the simulator for each match
+      --listen                       Listen for HTTP requests or supported log triggers and run the simulator for each match (not supported by cron)
```
timothyF95, line 33: 👍

timothyF95 (approved):

PR #502, opened 2026-06-23
TS solana binding generator

ernest-nowacki:
## TypeScript binding generator

`cmd/generate-bindings/solana/tsbindgen.go` (in the CLI repo) generates two files from an Anchor IDL:

- A **binding class** (`<ProgramName>.ts`) that:
  - Embeds the full IDL as a typed const.
  - Derives `@solana/codecs` codec expressions for every IDL type (structs, scalar enums, options, vecs, fixed arrays, all primitive types including pubkeys).
  - Sorts types so each codec const only references already-declared codecs.
  - Generates `writeReport`, `writeReportFromBorshEncodedVec`, `writeReportFrom<Struct>`, and `writeReportFrom<Struct>s` methods for every struct type.
  - Generates account and event decoders with discriminator validation (`decode<Name>Account`, `decode<Name>Event`).
- A **mock file** (`<ProgramName>_mock.ts`) that wires the binding class against `addSolanaContractMock` for use in unit tests.

## For the end user

The binding command will use the path `contracts/solana/src/idl/` by default when `--idl` is not specified. So user workflow directory structure would look like this:

```
<project-root>/
  contracts/
    solana/
      src/
        idl/          ← drop your Anchor IDL .json files here
        generated/    ← Go bindings written here
      ts/
        generated/    ← TypeScript bindings written here
```

Command scans every `*.json` in the idl directory and generates one binding class per file. You can also pass a single file with `--idl path/to/contract.json`. Language is auto-detected from whether the project contains `.go` or `.ts` files, or override with `--language typescript`.


timothyF95 (approved):
lgtm

PR #506, opened 2026-06-23
Add private-testnet-quartzite support

amit-momin:


go.mod
```diff
@@ -28,17 +28,17 @@ require (
...
 	github.com/smartcontractkit/chainlink-evm/gethwrappers v0.0.0-20260512150409-b4068bf735e6
-	github.com/smartcontractkit/chainlink-protos/cre/go v0.0.0-20260616232143-9a69bb7e7ebf
+	github.com/smartcontractkit/chainlink-protos/cre/go v0.0.0-20260623200841-e0322b819f62
 	github.com/smartcontractkit/chainlink-protos/workflows/go v0.0.0-20260528221400-84746b70eeeb
 	github.com/smartcontractkit/chainlink-testing-framework/seth v1.51.5
 	github.com/smartcontractkit/chainlink/deployment v0.0.0-20260521170940-67f9a4b233f8
 	github.com/smartcontractkit/chainlink/v2 v2.29.1-cre-beta.0.0.20260618123253-f1ee915b9a56
-	github.com/smartcontractkit/cre-sdk-go v1.12.0
-	github.com/smartcontractkit/cre-sdk-go/capabilities/blockchain/evm v1.0.0-beta.12
+	github.com/smartcontractkit/cre-sdk-go v1.13.0
```
anirudhwarrier, line 40: @amit-momin shouldn't this be v1.14.0?

anirudhwarrier (changes requested):

go.mod
amit-momin: Ideally it should be but the issue is the `chainlink-protos/cre/go` dependency which has to be a commit from the [capabilities-development](https://github.com/smartcontractkit/chainlink-protos/commits/capabilities-development/) branch. We have rules in the repo that requires us to push any changes to the cre package to the cap-dev branch then cherry-pick it to main. Since the 1.14.0 tag is on the commit in main, the commit in cap-dev is on the previous tag which is why this is only updating to 1.13.0 here.

amit-momin: The reason we have to use a cap-dev commit for `chainlink-protos/cre/go` is actually because of the `chainlink/v2` dependency. Seems like there's some code in core that relies on the cap-dev branch instead of main. So if I try to use a main commit, the CI will break.

anirudhwarrier (approved):

PR #523, opened 2026-07-08
Isolate test env process to prevent user context overwrite

timothyF95:
Redirect CLI config writes to a temp directory via testutil.IsolateCLIHome() in NewTestConfig, NewGraphQLMockServerGetOrganization, and private-registry subprocess tests.

### DIFF
--- a/internal/testutil/cli_home.go
+++ b/internal/testutil/cli_home.go
@@ -1,0 +1,76 @@
+package testutil
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// IsolateCLIHome redirects CLI config writes (~/.cre) into a temp directory.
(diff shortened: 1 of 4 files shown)


internal/testutil/cli_home.go
```diff
@@ -0,0 +1,84 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+var defaultGoPath string
+
+func init() {
+	home, err := os.UserHomeDir()
+	if err != nil {
```
ibrajer, line 14: What if there's an error? Do we need to handle that?

ibrajer (approved):

internal/testutil/cli_home.go
timothyF95: Should be okay, os.UserHomeDir() is very unlikely to fail and this also only behaves the default/backup.

PR #543, opened 2026-07-23
Fail closed on unparseable semver during self-update

anirudhwarrier:
[DEVSVCS-5289](https://smartcontract-it.atlassian.net/browse/DEVSVCS-5289)
## Fail closed on unparseable semver during self-update

**Problem**: 
cre update previously proceeded with the update whenever it couldn't parse the current or latest version as semver (fail-open), allowing unintended downgrades or version jumps. Flagged by security review.

**Fix:**
- Run now fails closed when either version is unparseable, returning an error and suggesting --force.
- Added a --force/-f flag to restore the warn-and-proceed behavior when explicitly requested.
- Normal downgrade/already-latest short-circuit is unchanged when both versions parse.
- Tests: added coverage for fail-closed on unparseable latest (asserting no download attempted), unparseable current, and --force bypass.

### DIFF
--- a/cmd/update/update.go
+++ b/cmd/update/update.go
@@ -300,8 +300,9 @@ func replaceSelf(newBin string) error {
 	return os.Rename(newBin, self)
 }
 
-// Run accepts the currentVersion string
-func Run(currentVersion string) error {
+// Run accepts the currentVersion string and a force flag that overrides the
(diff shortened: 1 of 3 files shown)


timothyF95 (approved):
lgtm

PR #555, opened 2026-07-27
Added address name linking mandatory with a default value

ejacquier:
## Summary

When linking a web3 key via cre account link-key, the interactive "owner label" prompt previously accepted an empty value silently and only failed later, deep in the GraphQL call, with a confusing error. This makes the prompt mandatory up front and gives it a sensible default so most users can just press Enter.

- Mandatory input: the prompt now rejects empty submissions locally with a label is required, instead of silently proceeding with "" and failing later against the linking API.
- Smart default: the default label is derived from the authenticated user's email (the part before @), pulled from the JWT access token claims via a new Credentials.GetEmail() method. The prompt shows it inline — Provide a label for your owner address [emmanuel.jacquier]: — and pressing Enter without typing accepts it.
- Falls back to a plain, still-mandatory prompt (no default) when no email is available, e.g. under CRE_API_KEY auth, where there's no JWT to read a claim from.
- Added ui.WithDefaultValue and ui.WithValidate options to the shared ui.Input helper so other commands can opt into the same pattern.


timothyF95 (changes requested):
Do we need a default or can we err on empty?

timothyF95 (approved):

